PR #410, opened 2014-12-31
Add MTT_TEST_SID env var

miked-mellanox:


jsquyres:
Sure, looks fine.  Just curious -- what do you use this for?

miked-mellanox:
we run mtt with "random" parameter to randomize tests selection order.
The existing "variant" test attribute represents serial number inside "test run" mtt group and from report it is hard to distinguish who was the 1st to run, test osu variant=1 or test imb variant=1, one need to go over timestamps....

so, having this parameter, we add to "exec" param the following:

cp $0 $SCRATCH_DIR/test_reproduce_${MTT_TEST_SID}.sh to see the actual order of test execution and have an easy repro script ready if needed.

PR #411, opened 2015-01-06
add start/stop site-wide hooks

miked-mellanox:


miked-mellanox:
@jsquyres - please review

jsquyres:
@miked-mellanox Have a look at #412 -- I offer it as an alternative to this PR.  It provides a foundation for including more hooks over time.

jsquyres:
In light of the conversation on #412, and combined with the good point that you want to set global hooks (without editing everyone's individual MTT .ini files), how about if we do something OMPI-MCA-param-like?  I think this could be a simple extension of your original idea. Specifically:

* Have a default funclet_files value that points to a .pm file in the MTT tree (i.e., if the user overrides the value, great, but otherwise it points to the .pm file in the MTT tree by default).
* The default funclet_files .pm file can have template (empty) hook step functions.
* When a step is to be called, it can check for the existence of the hook function.  If it exists, it is called.  If it does not exist, it is skipped (I think this happens already, actually).

This would allow a user to still specify their own funclet_files value and have their own hooks.  Or they can not specify their own funclet_files value and use the system-defined hooks.

Or, if you have users who are using their own funclet_files and still want to use the system hooks, then perhaps we can introduce a new field such as hook_file (or whatever) that behaves the same way: if not specified by the user, the global one is used.

Thoughts?

miked-mellanox:
wrapping it as a funclet will limit hook ability to read mtt.exe internals (globals, func locals  and init time variables).
I think we need access to mtt.exe internals and variables for hook to be useful.

miked-mellanox:
@jsquyres - could you please review

jsquyres:
How does putting perl in .pm files limit access to MTT internals?

Also, you aren't restricted to just MTT funclets.  Your .pm file can include put step functions in there that get called at relevant points (e.g., during the startup, similar to what you identified in #412).

jsquyres:
@miked-mellanox please review

miked-mellanox:
You cannot access script global variable from Hooks.pm module, please see example

``` shell

$cat aa.pl
#!/usr/bin/perl

use strict;
use lib './lib';

my $test_var="hello";

if(eval "require MTTHooks;") {
    print "Using Hooks\n";
    on_mtt_start();
} else {
    print "No hooks for you\n";
}

$cat lib/MTTHooks.pm
use strict;
use Exporter;
use vars qw($VERSION @ISA @EXPORT);

$VERSION    = 1.00;
@ISA        = qw(Exporter);
@EXPORT     = qw(on_mtt_start on_mtt_stop);

#sub on_mtt_start  { print "Hook ON for $test_var\n"}
sub on_mtt_start  { print "Hook ON\n"}
sub on_mtt_stop   { print "Hook OFF\n"}

1;

$./aa.pl
Using Hooks
Hook ON

$cat lib/MTTHooks.pm
use strict;
use Exporter;
use vars qw($VERSION @ISA @EXPORT);

$VERSION    = 1.00;
@ISA        = qw(Exporter);
@EXPORT     = qw(on_mtt_start on_mtt_stop);

sub on_mtt_start  { print "Hook ON for $test_var\n"}
#sub on_mtt_start  { print "Hook ON\n"}
sub on_mtt_stop   { print "Hook OFF\n"}

1;

$./aa.pl
No hooks for you
```

miked-mellanox:
@jsquyres - could you please review.

miked-mellanox:
@jsquyres - ping

miked-mellanox:
@jsquyres - could you please review

jsquyres:
Sorry for the delay.  

Here's an example showing a hook function accessing a global, and main accessing a variable from the hooks file:

MTTHooks.pm:
```perl
package MTTHooks;

use strict;

our $hook_var = "value of HOOK var";

sub func  {
    print "Hook func: hook var = $hook_var\n";
    print "Hook func: global var = $main::global_var\n";
}

1;
```

And main.pl:
```perl
#!/usr/bin/perl

use strict;
use lib "./lib";

our $global_var = "value of GLOBAL var";

if (eval "require MTTHooks;") {
    print "Using Hooks\n";
    MTTHooks::func();
    print "Back in main, hook var = $MTTHooks::hook_var\n";
    print "Back in main, global var = $global_var\n";
} else {
    print "No hooks for you\n";
}
```

When you run it:

```shell:
$ ./foo.pl
Using Hooks
Hook func: hook var = value of HOOK var
Hook func: global var = value of GLOBAL var
Back in main, hook var = value of HOOK var
Back in main, global var = value of GLOBAL var
```

miked-mellanox:
yep, this is correct, but will not serve the purpose:

* global var can be accessed only if they were defined with *our* keyword (in mtt.exe globals defined with *my*)

miked-mellanox:
@jsquyres - could you please comment.

jsquyres:
Not being able to change "my" variables is actually a *good* thing -- it's good protection.

Why not change the specific global variables that you want to change to be "our"?  That guarantees that you can only change specific, deliberate variables.

miked-mellanox:
ok. so will s/my/our/g for some variables which we plan to use inside the hooks.

jsquyres:
I think that's ok, yeah.  I.e.:

* We expand on an existing concept for hook files (in a very OMPI-like way)
* We expand the existing core to deliberately allow modifying of globals where needed

PR #412, opened 2015-01-06
hooks: add new hooks functionality to the mtt client

jsquyres:
In the INI file, you can define a hook_file and hook_prefix in the [MTT] section.  The hook_file is a .pm file that is required during the client startup.  The hook_prefix is the package name prefix of all the hook functions.

Two hook functions currently exist, but more can be easily added:

* mtt_hook_init
* mtt_hook_finalize

If a given hook function does not exist, it will be skipped (i.e., you do not have to define every hook in your hook file).

Here's an example Cisco_hooks.pm file:

```perl
package Cisco_hooks;

use strict;

sub mtt_hook_init {
    # Do something interesting here
}

1;
```

Notice that it only provides the mtt_hook_init function; the mtt client will detect that the finalize hook is not provided, and will silently/safely skip it.

You can tell the MTT client to use the above file by putting the following in the INI file:

```ini
hook_file = /path/to/Cisco_hooks.pm
hook_prefix = Cisco_hooks::
```

### DIFF
--- a/client/mtt
+++ b/client/mtt
@@ -106,6 +106,7 @@ use MTT::INI;
 use MTT::Reporter;
 use MTT::Defaults;
 use MTT::Globals;
+use MTT::Hooks;
 use MTT::FindProgram;
 use MTT::Trim;
 use MTT::DoCommand;
@@ -521,7 +522,10 @@ foreach my $hash (@ini_list) {
         next;
     }
 
-	
+    # Run the init hook, if it exists
+    MTT::Hooks::init($ini);
+    MTT::Hooks::invoke("mtt_hook_init");
+
     # Examine the [MTT] global defaults section.  We've already
     # determined the scratch_root, so pass it in so that it gets set
     # in the Globals hash.
@@ -649,6 +653,10 @@ foreach my $hash (@ini_list) {
         Verbose("Deleting fast scratch tree: $fast_scratch_arg\n");
         MTT::DoCommand::Cmd(1, "rm -rf $fast_scratch_arg");
     }
+
+    # Run the finalize hook, if it exists
+    MTT::Hooks::invoke("mtt_hook_finalize");
+    MTT::Hooks::finalize();
 }
 
 # That's it!
--- a/lib/MTT/Hooks.pm
+++ b/lib/MTT/Hooks.pm
@@ -1,0 +1,84 @@
+#!/usr/bin/env perl
+#
+# Copyright (c) 2006-2010 Cisco Systems, Inc.  All rights reserved.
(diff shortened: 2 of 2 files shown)


miked-mellanox:
The ini file aleady supports mtt:before_mtt_start_exec mtt:after_mtt_start_exec keywords which are resolved during mtt start/stop sequence.

The proposed mechanism allows to call start/stop hooks w/o modifying ini file and can be used as site-wide configuration/hooks for any user/ini file. (we have many users running mtt with different configurations and on different clusters but using same mtt.exe)

The use case for it - any mtt execution can submit its scratch dir and runtime parameters to some internal web blog system for easy html results navigation on per cluster/project/user/configuration basis.

jsquyres:
I think you're referring to the before_any and after_all keywords, which only refer to running something before any tests are run and after all tests are run.  These are quite different than the hook points you created in #411 (right during MTT init and finalize).  

#411 also uses a strange way to load perl code into the running process (cat and eval) that can be unsafe, depending on the contents of the sourced file.  This PR uses "require", which does a bunch of error checking, etc.  It also provides for hooks at specific points, and is easily extensible (just call MTT::Hooks::invoke("...")).

That being said, I see your use case -- site-wide defaults.  I agree that this is a good thing.  This is a new concept for the MTT client.  I'm not sure that inserting a .pl file in the middle of the git tree is the Right way to do this.

Hmm.

miked-mellanox:
how this pr differs from:
https://github.com/open-mpi/mtt/blob/master/client/mtt#L550
and
https://github.com/open-mpi/mtt/blob/master/client/mtt#L634

re #411 - the current implementation is as safe as any other code injection technique (in bash, source file, ...) but it allows to get easy access to mtt internals for intercept&take action from the hook.

eval will  fails (w/o failing mtt) if hook is not perl. this is kind of "live patching" mechanism.

miked-mellanox:
@jsquyres could you please review

jsquyres:
Oops!  Sorry for not responding for so long :-( -- thanks for the ping...

Err... ya, you're right.  The run_step stuff is already there.  So this PR is probably kinda moot.

But then #411 should be moot, too, right?  I.e., you can have a funclet_files add a .pm that adds a function that can be invoked via the run_step mechanism, right?

miked-mellanox:
yes, but i don`t want to edit many files in a lot of placec belonging to different users (mtt is a killer up here). I would like to use a fact that all users using same mtt.exe file and hook it in the central place.

jsquyres:
@miked-mellanox I'm going to close this (since it's moot) and move the discussion back to #411.

PR #415, opened 2015-04-01
HTMLReport: remove $domain

miked-mellanox:
it is not needed, because URL can ve relative and web server will add $url_base
automatically.

w/o $domain html files can be portable


miked-mellanox:
@jsquyres - please review

jsquyres:
:+1:

PR #417, opened 2015-07-21
MPI::OMPI::find_findings correctly handle quoted bindings

ggouaillardet:
until v2.x :
$ ompi_info -all --parseable | grep ^bindings:use_mpi:
bindings:use_mpi:yes (limited: overloading)

but since v2.x :
$ ompi_info -all --parseable | grep ^bindings:use_mpi:
bindings:use_mpi:"yes (limited: overloading)"

so we need to handle the optional double quote before yes...


ggouaillardet:
@jsquyres could you please review this ?
currently, the *_usempi test from the trivial test suite are not an with master and v2.x
(if you look at the absoft mtt results, there are 30 tests for v1.8 but only 18 for master)

jsquyres:
Sorry for the delay -- this came in while I was on vacation, and I just found a whole trove of emails from vacation that I never handled (!).  :frowning:

PR #418, opened 2015-09-23
Updating the link to OMPITesting page on README

kjtanaka:
The link to OMPITesting on the quick start section of README should be directed to mtt/wiki/OMPITesting on Github?

Thanks,


jjhursey:
Yes, it should. Good catch! Thanks for the patch.

PR #421, opened 2016-01-27
correctly retrieve fortran bindings information for OpenMPI 1.7 and hi…

ggouaillardet:
…gher

in MTT::MPI::Install::Analyze::OMPI::Install

### DIFF
--- a/lib/MTT/MPI/Install/Analyze/OMPI.pm
+++ b/lib/MTT/MPI/Install/Analyze/OMPI.pm
@@ -4,6 +4,8 @@
 #                         All rights reserved.
 # Copyright (c) 2006-2008 Cisco Systems, Inc.  All rights reserved.
 # Copyright (c) 2007      Sun Microsystems, Inc.  All rights reserved.
+# Copyright (c) 2016      Research Organization for Information Science
+#                         and Technology (RIST). All rights reserved.
 # $COPYRIGHT$
 # 
 # Additional copyrights may follow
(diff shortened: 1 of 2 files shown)


ggouaillardet:
@jsquyres @rhc54 i found that when using the ```AlreadyInstalled``` module with OpenMPI master and the trivial test suite
fortran tests were simply not built because the fortran bindings were reported missing

miked-mellanox:
:+1:

lib/MTT/MPI/Install/Analyze/OMPI.pm
```diff
@@ -46,13 +48,33 @@ sub Install {
...
-    Debug("Have F77 bindings: $ret->{f77_bindings}\n"); 
-    $ret->{f90_bindings} = &{$func}($ret->{bindir}, $ret->{libdir}, "f90");
-    Debug("Have F90 bindings: $ret->{f90_bindings}\n"); 
+
+    # OMPI 1.7 (and higher) refer to "bindings:mpif.h".  Prior
+    # versions refer to "bindings:f77".
+    my $tmp;
+    $tmp = &{$func}($ret->{bindir}, $ret->{libdir}, "f77");
+    $tmp = &{$func}($ret->{bindir}, $ret->{libdir}, "mpif.h")
+        if (!$tmp);
```
jsquyres: Should this be `if (!defined($tmp))`?

lib/MTT/MPI/Install/Analyze/OMPI.pm
```diff
@@ -46,13 +48,33 @@ sub Install {
...
+    $ret->{mpifh_bindings} = $ret->{f77_bindings} = $tmp;
+    Debug("Have mpif.h bindings: $ret->{mpifh_bindings}\n"); 
+
+    # OMPI 1.7 (and higher) refer to "bindings:use_mpi".  Prior
+    # versions refer to "bindings:f90".
+    $tmp = &{$func}($ret->{bindir}, $ret->{libdir}, "f90");
+    $tmp = &{$func}($ret->{bindir}, $ret->{libdir}, "use_mpi")
+        if (!$tmp);
```
jsquyres: Similar to above: `if (!defined($tmp))`

ggouaillardet:
This is a copy/paste from https://github.com/open-mpi/mtt/blob/master/lib/MTT/MPI/Install/OMPI.pm

```perl
sub find_bindings {
    my ($bindir, $libdir, $lang) = @_;

    my $file = _run_ompi_info($bindir, $libdir, "^bindings:$lang:");
    return ($file->[0] =~ /^bindings:${lang}:(")?yes/) ? "1" : "0";
}
```
find_bindings returns "0" or "1" so I think the code is currently correct

jsquyres:
@ggouaillardet Fair enough. 

FWIW: I prefer a defensive programming style -- unexpectedly undefined perl vars can lead to difficult-to-find bugs later.  As such, I like to guard for them whenever possible.

ggouaillardet:
so the correct statement should be
```perl
... if (!defined($tmp) || !$tmp);
```
or maybe in a more compact and perlish form
```perl
... unless($tmp);
```

I do not mind updating the PR to avoid bugs in the future

jsquyres:
Yes, I think you're right: `if (!defined($tmp) || !$tmp)`.

Cool; thanks.

jsquyres:
Wait, I missed your `unless` form... does `unless` also check for `!defined`?  I never keep that stuff straight.  As you can probably tell, I write perl very much like I write C.  :smile:

ggouaillardet:
@jsquyres done
finally, i replaced ```if (!$tmp)``` with ```unless(defined($tmp) && $tmp)```
omitting the ```defined()``` test will likely generate a warning if the variable is not defined

jsquyres:
:+1:

PR #422, opened 2016-01-27
add the trivial_tests_f08flags config option in the Trivial Build mod…

ggouaillardet:
…ule.

trivial_tests_f08flags is used to build f08 programs when defined, and
trivial_tests_fflags is used otherwise.

### DIFF
--- a/lib/MTT/Test/Build/Trivial.pm
+++ b/lib/MTT/Test/Build/Trivial.pm
@@ -4,6 +4,8 @@
 #                         All rights reserved.
 # Copyright (c) 2006-2012 Cisco Systems, Inc.  All rights reserved.
 # Copyright (c) 2007      Sun Microsystems, Inc.  All rights reserved.
+# Copyright (c) 2016      Research Organization for Information Science
+#                         and Technology (RIST). All rights reserved.
 # $COPYRIGHT$
 # 
 # Additional copyrights may follow
@@ -56,6 +58,12 @@ sub Build {
                        "trivial_tests_cflags");
     my $fflags = Value($ini, $config->{full_section_name}, 
                        "trivial_tests_fflags");
+    my $f77flags = Value($ini, $config->{full_section_name}, 
+                       "trivial_tests_f77flags");
+    my $f90flags = Value($ini, $config->{full_section_name}, 
+                       "trivial_tests_f90flags");
+    my $f08flags = Value($ini, $config->{full_section_name}, 
+                       "trivial_tests_f08flags");
     my $languages = Value($ini, $config->{full_section_name}, 
                        "trivial_tests_languages");
     my $mpicc  = Value($ini, $config->{full_section_name}, 
@@ -132,10 +140,10 @@ sub Build {
 
     if ($mpi_install->{mpifh_bindings} and $languages_hash->{"mpifh"}) {
         Debug("Test compile/link sample mpif.h MPI application\n");
-        $x = _do_compile("$mpif77 $fflags", "hello_mpifh.f", "hello_mpifh");
+        $x = _do_compile("$mpif77 $fflags $f77flags", "hello_mpifh.f", "hello_mpifh");
         return $x
             if (defined($x));
-        $x = _do_compile("$mpif77 $fflags", "ring_mpifh.f", "ring_mpifh");
+        $x = _do_compile("$mpif77 $fflags $f77flags", "ring_mpifh.f", "ring_mpifh");
         return $x
             if (defined($x));
     } else {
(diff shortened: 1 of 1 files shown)


ggouaillardet:
@jsquyres this is related to open-mpi/ompi#1327

just add
```
trivial_tests_f08flags = -std=f2008
```
in the ```[Test build: trivial]```  section, and MTT will fail building the trivial test suite with a f08 capable gfortran

note that is not perfect because it is only valid for gnu compiler.

miked-mellanox:
:+1:

lib/MTT/Test/Build/Trivial.pm
```diff
@@ -56,6 +58,8 @@ sub Build {
                        "trivial_tests_cflags");
     my $fflags = Value($ini, $config->{full_section_name}, 
                        "trivial_tests_fflags");
+    my $f08flags = Value($ini, $config->{full_section_name}, 
+                       "trivial_tests_f08flags");
```
jsquyres: If we do this, we should probably do it properly: add mpifh_flags, usempi_flags, and usempif08_flags.

We can probably leave fflags -- those can be flags that apply to all fortran programs, but then you can use the individual flags vars to specify flags for specific bindings.

ggouaillardet:
ok, will do

lib/MTT/Test/Build/Trivial.pm
```diff
@@ -161,11 +165,12 @@ sub Build {
     # linking simple "use mpi_f08" applications
 
     if ($mpi_install->{usempif08_bindings} and $languages_hash->{"usempif08"}) {
+        $f08flags = $fflags if (!defined($f08flags));
         Debug("Test compile/link sample \"use mpi_f08\" MPI application\n");
-        $x = _do_compile("$mpif08 $fflags", "hello_usempif08.f90", "hello_usempif08");
+        $x = _do_compile("$mpif08 $f08flags", "hello_usempif08.f90", "hello_usempif08");
```
jsquyres: If we're not going to get rid of fflags, then you should have *both* `$fflags` and `$usempif08_flags` in there:

```perl
$x = _do_compile("$mpif08 $fflags $usempif08_flags", ...
```

lib/MTT/Test/Build/Trivial.pm
```diff
@@ -161,11 +165,12 @@ sub Build {
...
         return $x
             if (defined($x));
-        $x = _do_compile("$mpif08 $fflags", "ring_usempif08.f90", "ring_usempif08");
+        $x = _do_compile("$mpif08 $f08flags", "ring_usempif08.f90", "ring_usempif08");
```
jsquyres: Same comment as above about fflags and usempif08_flags.

lib/MTT/Test/Build/Trivial.pm
ggouaillardet: so what is the spec of usempif08_flags ?
does it override fflags ? or should it be appended to fflags for f08 programs only ?

jsquyres: I think the two should be used in conjunction with each other: general fortran flags (e.g., if you need to pass something to *all* fortran compilations), and per-bindings flags (e.g., if you need to pass something to just the `use mpi_f08` tests, as this PR was created for).

ggouaillardet: ok, will do that

ggouaillardet:
@jsquyres done
i noticed c++ uses ```trivial_tests_cflags```, should it used ```trivial_tests_cxxflags``` instead ?

lib/MTT/Test/Build/Trivial.pm
```diff
@@ -56,6 +58,12 @@ sub Build {
                        "trivial_tests_cflags");
     my $fflags = Value($ini, $config->{full_section_name}, 
                        "trivial_tests_fflags");
+    my $f77flags = Value($ini, $config->{full_section_name}, 
+                       "trivial_tests_f77flags");
+    my $f90flags = Value($ini, $config->{full_section_name}, 
+                       "trivial_tests_f90flags");
+    my $f08flags = Value($ini, $config->{full_section_name}, 
+                       "trivial_tests_f08flags");
```
jsquyres: I'm sorry -- these names are wrong.  They are not the "F77 bindings", "F90 bindings", and "F08 bindings", they are the "mpif.h interface", the "'use mpi' interface", and the "'use mpi_f08' interface".

For example: the mpif.h interface does not conform to the Fortran 77 standard.  Nor does the "use mpi" interface conform to Fortran 90.

ggouaillardet: I was just being symmetric with $mpif77 $mpif90 and $mpif08
shall I replace these too ? or is it fine to be somehow inconsistent ?

jsquyres: Might as well change them while we're mucking around in here...

Sorry; you're getting caught up in my type-A personality, and the fact that I'm on the Fortran WG in the MPI Forum (i.e., I'm on a crusade to fix everyone's nomenclature, because the use of "F77", "F90", and "F08" is just wrong.  And yes, even Past Jeff got it wrong in this MTT code... :smile: ).

If you want to just commit this one update for the new flags, I'll be happy to do the 2nd fix of updating the rest of the $mpif77 / $mpif90 / $mpif08 stuff (I didn't look in the git logs, but it was almost certainly me who wrote that stuff in the first place!).

jjhursey:
What's the state of this PR? It's been out here for a while.

jjhursey:
@ggouaillardet @jsquyres Any updates on this PR?

jsquyres:
Not from me -- @ggouaillardet?

jjhursey:
I'm going to close this PR, but feel free to reopen as needed.

PR #423, opened 2016-02-15
https support :

ggouaillardet:
- build full URL with the appropriate protocol (e.g. https vs http)
- update all cron scripts to use https://mtt.open-mpi.org


ggouaillardet:
@jsquyres can you please review this ?

currently, when clicking on a button in the MTT web ui, a warning is issued since a form is being submitted to a non secure url (e.g. http://mtt.open-mpi.org/) from a secure url (e.g. https://mtt.open-mpi.org) and this is really annoying.

note i automatically detect whether the web site is using SSL 
```php
if (array_key_exists("HTTPS", $_SERVER) && $_SERVER["HTTPS"] == "on") {
    $GLOBALS["protocol"] = "https";
} else {
    $GLOBALS["protocol"] = "http";
}
```
and i do not know whether this is standard or good for apache/mod_ssl only.

out of curiosity, is there any reason why we explicitly build full URI
for example
```php
$self = $GLOBALS["protocol"].'://' . $_SERVER['SERVER_NAME'] . $_SERVER['SCRIPT_NAME'];
```
my point is, could we simply do (assuming the path starts with a ```/```)
```php
$self = $_SERVER['SCRIPT_NAME'];
```

jsquyres:
@ggouaillardet Thanks for this.

Yes, there are many bad things about the reporter PHP.  It's effectively spaghetti code -- it's unmaintainable.  :frowning: Josh Hursey did some work that may be able to replace it, and Intel is doing some work that may replace it, too.  We'll see how it works out...

jjhursey:
That looks good. Thanks!

The current MTT installation at mtt.open-mpi.org is not tracking the git repo at the moment. I'll take a ticket to fix that this week so this fix is rolled out properly.

PR #425, opened 2016-02-18
mtt global config - add global time quota per test run

miked-mellanox:
mtt mtt:per_test_timeout=10m

useful to override test defined timeout with global value.

### DIFF
--- a/lib/MTT/Test/RunEngine.pm
+++ b/lib/MTT/Test/RunEngine.pm
@@ -444,9 +444,15 @@ sub _run_one_test {
     $ENV{MTT_TEST_NP} = $MTT::Test::Run::test_np;
     _run_step($mpi_details, "before_each");
 
+    my $pertest_timeout = MTT::Values::Value($ini, 'mtt', 'per_test_timeout_global_override');
     my $timeout = MTT::Values::EvaluateString($run->{timeout}, $ini, $test_run_full_name);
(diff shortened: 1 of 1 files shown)


jsquyres:
@miked-mellanox I'm not sure I understand: it looks like setting this global value overrides *all* individual test timeouts.  Shouldn't it be the other way around?  I.e., set a global timeout, and then individual tests can override that global value with their own value?

jsquyres:
@miked-mellanox Also, regardless of which way this goes, if this gets merged, can you update the MTT global vars documentation / listing in the wiki so that people can know that this variable is available?  Thanks.

miked-mellanox:
@jsquyres - sure, will update wiki.

The use-case in mind  is that:
* mtt.ini has a lot of tests, with [run test].timeout set and tuned for large system scale
* now Im running it on small system and when test hangs - it timeouts for a long time
* what is needed - a way to "force" timeout for all, regardless of local timeout.

jsquyres:
@miked-mellanox Ok, good enough.  How about a stronger variable name, then?  I.e., instead of `per_test_timeout`, how about `per_test_timeout_global_override`, or something like that?

miked-mellanox:
good idea, will change. thx

miked-mellanox:
updated wiki

PR #426, opened 2016-03-14
lib/SCM: Remove leading/trailing spaces from scm_url

jjhursey:
If you add extra spaces at the end if the scm_url then chdir in ``_git_identify_n`` will throw the error:
 ``fatal: Not a git repository ...``
Seems to only happen on Ubuntu 14.04 (did not have the problem on RHEL 7.2).

 * Minor additional fix to change a debug 'print' to a protected Debug

### DIFF
--- a/lib/MTT/Util.pm
+++ b/lib/MTT/Util.pm
@@ -2,6 +2,7 @@
 #
 # Copyright (c) 2007-2008 Cisco, Inc.  All rights reserved.
 # Copyright (c) 2007 Sun Microsystems, Inc.  All rights reserved.
+# Copyright (c) 2016      IBM Corporation.  All rights reserved.
 # $COPYRIGHT$
 # 
 # Additional copyrights may follow
@@ -29,6 +30,7 @@ use base qw(Exporter);
              convert_time_to_human
              get_array_ref
              merge_hashes
+             trim_spaces
 );
 
 use MTT::Globals;
@@ -489,5 +491,14 @@ sub term_handler{
     }
 	Verbose("All is done, exiting.\n");
 	exit(0);
+}
+
+#--------------------------------------------------------------------------
+
+# Trim the leading and trailing whitespaces
+sub trim_spaces {
+    my ($str) = @_;
+    $str =~ s/^\s+|\s+$//g;
+    return $str;
 }
 1;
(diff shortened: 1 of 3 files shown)


jsquyres:
:+1:

SACHIN-BAPPALIGE:
$str =~ s/^\s+|\s+$//g
 where (\s+)  ===> 1 or more white spaces 
-------------------------------------
How (\s*) is better than  (\s+) ?
$str =~ s/^\s*|\s*$//g
-------------------------------------

jsquyres:
@SACHIN-BAPPALIGE Github pro tip: enclose code in single ticks so that it isn't accidentally rendered as wiki markup.  For example:

How `(\s* )` is better than `(\s+)` in trim function?
1) `$str =~ s/^\s*|\s *$//g`
2) `$str =~ s/^\s+|\s+$//g`
where `(\s+)` ===> 1 or more white spaces

Also note that you can click on the "Files changed" tab on the web interface in this PR and then click on the "+" right on the line in question of the code proposed by this PR, and add a comment (like your relevant question) right there.  It's pretty handy.

lib/MTT/Util.pm
```diff
@@ -490,4 +492,13 @@ sub term_handler{
 	Verbose("All is done, exiting.\n");
 	exit(0);
 }
+
+#--------------------------------------------------------------------------
+
+# Trim the leading and trailing whitespaces
+sub trim_spaces {
+    my ($str) = @_;
+    $str =~ s/^\s*|\s*$//g;
```
jsquyres: @jjhursey I think @SACHIN-BAPPALIGE is asking about this line.

jjhursey: In my mind they will do the same thing here. Would ``\s+`` be better form/style than ``\s*`` in this case?

jjhursey:
As a broader note, the problem is that ``chdir`` was failing if there is whitespace at the end of the string. This commit just fixes the one instance where this is causing an issue.

Should we have a replacement for ``chdir`` that always *trims* the string before passing it to ``chdir``.

jsquyres:
I think it's ok to *not* trim in `chdir`, because you might have a crazy case where a directory name *does* begin or end in whitespace (which is legal in the filesystem).

lib/MTT/Util.pm
SACHIN-BAPPALIGE: Thanks Jeff and Josh  for your information. When I printed a execution time for both cases, there was a little variation.

jjhursey: So the \+ is approx. 600 microseconds faster than the \*. I'll update the patch.

jsquyres: Fair enough.  But don't forget that that's 4-5 (or more!) orders of magnitude less than the length of a typical MTT client run.  Optimizations are always good, but we probably don't need to optimize for milisecond gains deep down inside MTT.  It's Perl, after all.  :smile:

PR #427, opened 2016-03-29
Create CherryPy Submit server & updated client submit script

jjhursey:
* Create a CherryPy server
   - Provides a JSON REST interface for submitting results
 * Create a Perl Reporter Module (`MTTStorage`) to submit to this interface
 * Update the Python Reporter Module (`IUDatabase`) to submit to this interface

API is documented here:
  https://github.com/jjhursey/mtt-modern/wiki/REST-Interface-(Submit-Side)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,8 @@ server/php/submit/php_errors.log
 db.sqlite3
 */mttscratch
 
+server/php/cherrypy/bin/mtt_server.cfg
+server/php/cherrypy/.htaccess
+server/php/cherrypy/src/env
+server/php/cherrypy/log/
+server/php/cherrypy/src/.mtt_server.pid
--- a/lib/MTT/MPI/Install.pm
+++ b/lib/MTT/MPI/Install.pm
@@ -825,6 +825,11 @@ sub _do_install {
             $ret->{$k} = $serials->{$module}->{$k};
         }
 
+        $module = "MTTStorage";
+        foreach my $k (keys %{$serials->{$module}}) {
+            $ret->{$k} = $serials->{$module}->{$k};
+        }
+
         MTT::MPI::SaveInstalls($install_base,
             $MTT::Globals::Internals->{mpi_get_name},
             $mpi_get->{version},
--- a/lib/MTT/Reporter/MTTStorage.pm
+++ b/lib/MTT/Reporter/MTTStorage.pm
@@ -1,0 +1,653 @@
+#!/usr/bin/env perl
+#
+# Copyright (c) 2005-2006 The Trustees of Indiana University.
+#                         All rights reserved.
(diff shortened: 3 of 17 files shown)


rhc54:
@noahv Just pointing Noah to your PR - please let us know how we can be of help!

jjhursey:
I think I'll need some help finishing the IUDatabase file. If you search for JJH in the IUDatabase file you will see a set of fields all set to `None`. We will need to figure out the best way to fill in these values based upon what the Python client has collected (we might need to track some more information).

```
shell$ grep -n JJH pylib/Stages/Reporter/IUDatabase.py
177:		# JJH Todo fill these fields in
255:		# JJH Todo fill these fields in
308:		# JJH Todo fill these fields in
```

I'll try to get some more cycles to work on this in the next few days, but if you have some cycles to pick away at it that would help. Once those fields are filled then then we should be all set.

I'm working on setting up the new cherrypy service on `mtt.open-mpi.org`. It needs some installs, so it might not be until early next week that the service will be active.

rhc54:
@jjhursey I see one problem - we don't record the final parameters used to execute the section. I'll update that so we do.

I'm not sure what some of these other fields mean - e.g., mpi_install_id? Is that a character string of the stage that installed the MPI being used?

Perhaps a phone call could help clarify those?

rhc54:
@jjhursey Okay, I pushed a change (open-mpi/mtt@54b2e7be42d4e82631e72a04b50784bd12970d68) that records the final stage options (i.e., the parsed combination of default options and provided input params) as a dictionary called 'options' in the log. This will contain much of the info you were asking for, so I think what is left is for us to decide on the meaning of the remaining fields.

jjhursey:
@rhc54  The `mpi_install_id` will be one of the parameters passed back by the server along with the `submit_id` (and some other ids). So those just need to be handed back to the server, if the client still has them. I should writeup some documentation on that interface.

I have the server running here:
   https://mtt.open-mpi.org/submit/cpy/api/
Use the same credentials that you would use for the normal IU submit database. It will submit to the IU database so you can see the results in the Reporter - although everything is marked as 'trial'.

I'll look at the change today, and see if I can't match up some of the fields. Then it would probably be good to have a phone call about any remaining items.

jjhursey:
I added some documentation on the fields here:
  https://github.com/jjhursey/mtt-modern/wiki/REST-Interface-(Submit-Side)

It includes example input/output.

jjhursey:
I would like to merge this into the master branch since most of the backend work is finished, along with the Perl client plugin. The Python plugin will require some iteration which is happening primarily on the master branch.

@rhc54 What do you think?

rhc54:
Please do! Much appreciated

PR #428, opened 2016-04-07
Add detection of compiler vendor and version

rhc54:
@jjhursey Could you take a look at this? I mirrored it on the code in Functions.pm, but I don't know the executable name for some of these compilers - i.e., which ones use "cc" versus family-specific names like "gcc". Heck, I don't even recognize many of these any more, and we may not really care about them all.


jjhursey:
:+1: I think that looks good.

Yeah, a bunch of those I don't recognize either, and we probably don't care to detect them any more.

This should give us the `compiler_name` and 'compiler_version` information that the server wants to know. You can pass the same information for the `MPI Install` and `Test Build` phases (it is _rare_ that anyone would want them to be different).

PR #429, opened 2016-04-12
whatami: Add ppc64le support, and alternative RHEL detection

jjhursey:
* Add support for pcc64le (Power8 arch)
* Add support for an alternative to detecting RHEL version information (we have a system where `/etc/issue` is not reliable)

### DIFF
--- a/client/whatami/whatami
+++ b/client/whatami/whatami
@@ -169,6 +169,9 @@ get_linux_type()
                 i386|i486|i586|i686)
                         hardware=ia32
                         ;;
+                ppc64le)
+                        hardware=ppc64le
+                        ;;
                 ppc64)
                         hardware=ppc64
                         ;;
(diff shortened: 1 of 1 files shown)


jjhursey:
@jsquyres @rhc54 Can one or both of you take a look at the RHEL detection component of this PR. It shouldn't case any problems, but I wanted another set of eyes on it before pushing it back.

It first checks for `/etc/redhat-release` before looking at `/etc/issue` for the RHEL information.

Notice also that the RHEL detection only captures the major version (`7` when it is `7.3`). This happens for RHEL 6 too. Is this something that should be fixed?

client/whatami/whatami
```diff
@@ -219,6 +222,11 @@ get_linux_type()
                         #distro=${distro_brand}_${distro_version}
                         distro=${distro_brand}
 
+                elif [ -f /etc/redhat-release ]; then
```
jjhursey: I want to add a further check to this statement to make sure it contains the correct string.

Something like:
```bash
elif [ -f /etc/redhat-release -a -n "`egrep 'Red Hat Enterprise Linux ([a-zA-Z]+) release [0-9]*' /etc/redhat-release`" ]; then
```

I'll make the update shortly.

jjhursey:
I think this update properly addresses my earlier concern. 👍

rhc54:
@jjhursey sorry I didn't get to this earlier - I think it looks fine.

PR #430, opened 2016-04-27
Funclets: add &test_mpirun_args()

jsquyres:
Allow Specify::Simple to pass tokens to &test_mpirun_args() (i.e., arguments to mpirun itself, not the test executable).  Its intent is to to do something like this:

```
exec = mpirun &test_mpirun_args() -np &test_np() ... &test_executable() &test_argv()
```

You set them in the Test Run section like this:

```
simple_dyn:tests = src/mpi2c++_dynamics_test
simple_dyn:mpirun_args = --oversubscribe --bind-to none
```

The `mpirun_args` clause will get returned from `&test_mpirun_args()`.

@jjhursey Please review.

### DIFF
--- a/lib/MTT/Test/RunEngine.pm
+++ b/lib/MTT/Test/RunEngine.pm
@@ -230,7 +230,7 @@ sub RunEngine {
         $MTT::Test::Run::test_executable_dir = $test_exe_dir;
         $MTT::Test::Run::test_executable_abspath = $test_exe_abs;
         $MTT::Test::Run::test_executable_basename = $test_exe_basename;
-
+        $MTT::Test::Run::test_mpirun_args = $run->{mpirun_args};
         $MTT::Test::Run::test_argv = $run->{argv};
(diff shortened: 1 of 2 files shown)


jjhursey:
👍 
This seems really useful for the dynamics cases. Can you update the wiki after this is merged in?

The `MPI Details` section would specify arguments for all `Test Run` sections, where the `&mpirun_args` would be additional arguments for a specific test run. Is it possible for a user to specify a default `&mpirun_args` that can be overridden by a specific test? (I believe so, but I don't recall what the INI syntax would be for that.)

jsquyres:
Yes, I'll update the wiki.

I don't think that there is syntax for a default output value of `&test_mpirun_args()` that can be overridden at any given test run section.

rhc54:
I'll try to add that in the Python client library - we do have the ability to override defaults pretty easily. I'm still trying to work my way thru the funclets and reproduce the capabilities in the Python area, so I'll add this to the list.

PR #434, opened 2016-05-27
Refactored environment modules plugin to use existing lmod python wra…

noahv:
…pper.  That wrapper uses python exec which sets the env module context in the pyclient parent context

which carries through to subsequent subprocess invocation contexts.
Added --env-module-cmd-path to allow user to point to lmod install if not set in LMOD_PKG.
Fixed a bug in the Autotools plugin exception handling which prevented running without the autogen phase.
Added module and module_unload to the Shell plugin options.  Also added status check for Shell plugin command execution.
Added module and module_unload to the SLURM plugin options.  Added SLURM plugin support for both load and unload of modules based on previous middleware or build stages.

### DIFF
--- a/pyclient/pymtt.py
+++ b/pyclient/pymtt.py
@@ -97,8 +97,8 @@ execGroup.add_option("--group-results", dest="submit_group_results", default=Tru
                      help="Report results from each test section as it is completed")
 execGroup.add_option("--default-make-options", dest="default_make_options", default="-j10",
                      help="Default options when running the \"make\" command")
-execGroup.add_option("--module-cmd", dest="module_cmd", default="module",
-                     help="Command used to load/unload a module")
+execGroup.add_option("--env-module-cmd-path", dest="env_module_cmd_path", default=None,
+                     help="Path to the lmod python env_modules_python.py wrapper")
 parser.add_option_group(execGroup)
 
 debugGroup = OptionGroup(parser, "Debug Options")
--- a/pyclient/pymtt.py
+++ b/pyclient/pymtt.py
@@ -97,8 +97,8 @@ execGroup.add_option("--group-results", dest="submit_group_results", default=Tru
                      help="Report results from each test section as it is completed")
 execGroup.add_option("--default-make-options", dest="default_make_options", default="-j10",
                      help="Default options when running the \"make\" command")
-execGroup.add_option("--env-module-cmd-path", dest="env_module_cmd_path", default=None,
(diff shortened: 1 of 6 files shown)


rhc54:
@noahv I'm a little disturbed that our module support appears to be tightly coupled to lmod, which I don't believe any MTT or OMPI community member uses. Can we leave this to be more general?

I'd like to see us retain support for modules in general, and only use lmod commands if lmod is present. Is that doable?

noahv:
Good question.  If other module packages support python we can look at making this plugin more flexible or implement individual plugins for each module package.

What module packages are being used by the community?

noahv:
Good question.  If other module packages support python we can look at making this plugin more flexible or implement individual plugins for each module package.

What module packages are being used by the community?

rhc54:
Everyone here uses the old "module" system - I can check to see about its python support. If it doesn't have it, we might be able to make it work anyway as I believe all it involves is getting python to "source" the script so the envars apply to the running code.

Hope you feel better soon!

rhc54:
I looked at the module system, and it also provides a python interface. Here is what it contains - should be easy to integrate into our client:

```python
import os, string

os.environ['MODULESHOME'] = '/home/common/tcl'

if os.environ.has_key('TCLSH'):
   TCLSH=os.environ['TCLSH']
else:
   if os.path.exists('/usr/bin/tclsh'):
      TCLSH="/usr/bin/tclsh"
   else:
      if os.path.exists('/bin/tclsh') :
         TCLSH="/bin/tclsh"
      else: 
         TCLSH=""

if not os.environ.has_key('MODULEPATH'):
        os.environ['MODULEPATH'] = '/mips/tools/freeware/modulefiles'

if not os.environ.has_key('LOADEDMODULES'):
        os.environ['LOADEDMODULES'] = '';

def module(command, *arguments):
        commands = os.popen('$TCLSH /home/common/tcl/modulecmd.tcl python %s %s' % (command, string.join(arguments))).read()
        exec commands
```

noahv:
Thanks.  Chest colds are no fun.

I will update the change set to support this as well.

  -- Noah

noahv:
I pushed an update to noahv:topic/env_modules_refactor that makes the support for modules more flexible.  The new usage for tcl modules would is "--env-module-wrapper=/home/common/tcl/init/python" if you installed the tcl module support to /home/common.

Thanks,

  -- Noah

rhc54:
kewl - thanks Noah!!

👍

PR #441, opened 2016-06-14
Uncovered some issues when trying to build and run from the OMPI github repo.

rhc54:
These changes fix that problem and move us closer to running cleanly. Still getting an error on the tests themselves, but need to further diagnose it.


rhc54:
@noahv Would you please take a look at this?

rhc54:
@noahv is working on some additional debug here, and will generate a PR against my branch when done. So I'll hold off and commit once he is done and we can bring it all over together

PR #447, opened 2016-07-01
Added files to create an RPM Package. (for Python files)

lonnystrunk:
Using the setup.py file with this command:
python setup.py sdist
to create the tarball from the list of files in MANIFEST.in,
you will have two files in dist/
mtt-{version}.tar.gz & mtt.spec

Those two files can be used to create the RPM. Either with OBS or an rpmbuild command.

This implementation will install only the folders and files specified in the MANIFEST.in file directly into the folder: /opt/mtt/
It is a relocatable package, so it can be installed wherever it is specified with the --prefix option.
It will not be installed into the default Python module location.

### DIFF
--- a/dist/mtt.spec
+++ b/dist/mtt.spec
@@ -1,0 +1,98 @@
+%define baseName mtt
+%define version 0.1
+%define install_path /opt/%{baseName}
+
+# Used to ignore files packaged automatically from sdist command
+# makes error become a warning when there are installed but unpackaged files
+%define _unpackaged_files_terminate_build 0
+
+Summary: MTT
+Name: %{baseName}
+Version: %{version} 
+Release: 1
+Source0: %{baseName}-%{version}.tar.gz
+License: BSD
+Group: Development/Libraries
+BuildRoot: %{_tmppath}/%{baseName}-%{version}-%{release}-buildroot
+Prefix: /opt 
+BuildArch: noarch
+Url: https://github.com/open-mpi/mtt
+
+# Use RPM's automatic dependency processing? (yes by default)
+#AutoReqProv: no
+
+%description
+This is the MPI Testing Tool (MTT) software package.  It is a
+standalone tool for testing the correctness and performance of
+arbitrary MPI implementations.
+
+%package common
+Summary: MTT Common Files
+Group: Development/Libraries
+%description common
+This is the MPI Testing Tool (MTT) software package for the common files.
+
+%package pyclient
+Summary: MTT PyClient
+Group: Development/Libraries
+Requires: mtt-common, python, python-virtualenv 
+%description pyclient
+This is the MPI Testing Tool (MTT) software package for the python client.
+
+%package server 
+Summary: MTT Server 
+Group: Development/Libraries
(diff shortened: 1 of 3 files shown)


jjhursey:
This only covers the python client as an RPM package, correct?

I can see this being useful for a site that just wanted to run MTT and submit results elsewhere. We should probably talk about how to make this package different than the 'sever' package.

jjhursey:
We chatted about this on the devel call today. A few notes:
 * We should have 2 RPMs
   - `mtt-pyclient` : Just the Python client to make it easier for a site to deploy/run the client without the weight of the server.
   - `mtt-server` : All of the server side MTT infrastructure (Issue #449)
 * Question about futurization of the RPM to work for both Python 2 and 3 (See PR #446). Virtualenv might be the solution here for user environment. Is this something that needs to be in the RPM or just user documentation?

lonnystrunk:
I think I found a good solution. I'll try and implement it in the next couple of days. 
http://www.rpm.org/max-rpm/s1-rpm-subpack-spec-file-changes.html

lonnystrunk:
Since we aren't going to use the RPM to install a python module, it makes more sense to not use the setup.py & MANIFEST.in files and just create the tarball of the directory and then have the spec file choose the files to include in the package. The spec file will create both RPMs: `mtt-pyclient` and `mtt-server`.

The command I used for creating the tarball from the git repo was:
`git archive --format=tar.gz --prefix=mtt-<version>/ <branchName> > dist/mtt-<version>.tar.gz`
You will have to make sure the versions match in the spec file for it to work correctly.

rhc54:
It looks to me like this should accomplish our desired objectives.

@jjhursey does the server look like it has everything it needs?

jjhursey:
That's fine for the server. Is there a means to require some dependencies? I'm thinking CherryPy and PostgreSQL / MySQL would be nice to require so they are available. Apache and PHP would also be in the list.

I'd like to see the Perl client included. It just needs the `client` and `lib` directories.

We should probably also include the `samples` directory when installing either of the client's.

lonnystrunk:
@rhc54 @jjhursey Would you be able to look at it again?
 I added the dependencies and fixed a couple of files that were giving me trouble when I tried to create the RPMs on OBS. I added all of the changes and the perlclient. I did just a basic install test but not a functionality test.

rhc54:
it looks right to me - @jjhursey ?

jjhursey:
That is looking good. The perl client has a number of libraries that it needs - I don't think we need to enumerate them all now. But maybe add the requirement for `LWP::UserAgent` (whichever package contains that) to get started then we can extend it as we start to use it.

lonnystrunk:
@jjhursey Added.

jjhursey:
This looks good to me. 👍 

It will probably need some refinement as we start working on clean deployments using the RPM (to see if we missed any deps). Should we bring this in now and iterate on the list, or iterate on the PR until we have it tested a bit more?

rhc54:
Let's bring it in now as it is an optional feature and thus won't impact anyone. Will be easier to iterate there

PR #448, opened 2016-07-03
Analyze: allow treating timeouts as failures

jsquyres:
Per request from the Open MPI community, add a parameter
"treat_timeouts_as_fail" to be set in Test Run sections that, upon
result of a timeout, will override that result and classify it as a
failure.

The rationale is that timeouts are not frequently examined in the web
reporter, and at least some kinds of test should treat timeouts as
failures.

@miked-mellanox @jjhursey


jjhursey:
This looks good to me. Thanks. 👍

PR #451, opened 2016-07-08
Added Environ plugin which enables setting environment variables.

noahv:

### DIFF
--- a/pylib/System/TestDef.py
+++ b/pylib/System/TestDef.py
@@ -438,6 +438,8 @@ class TestDef(object):
                 print("Test .ini file not found!: " + testFile)
                 sys.exit(1)
         self.config = configparser.ConfigParser()
+        # Set the config parser to make option names case sensitive.
+        self.config.optionxform = str
         self.config.read(self.log['inifiles'])
         for section in self.config.sections():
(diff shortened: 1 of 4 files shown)


rhc54:
I don't think that is quite correct - you really don't want to have to list every value in the module "options" area. All you really need is something like this:

```python
kvkeys = keyvals.key()
for kvkey in kvkeys:
    os.environ[kvkey] = keyvals[kvkey]
```

This way, every "key = value" pair will be pushed into the environment, not just the ones we hardcoded into the plugin.

Make sense?

noahv:
Sure, we can make it more flexible.   I will update the pull request.

Thanks,

  -- Noah

noahv:
@DebRez I added an easy two line fix for the io.StringIO.seek() instead of StringIO.StringIO.len().
Please take a peak.

Thanks,

  -- Noah

rhc54:
Looks fine to me

DebRez:
Looks good to me, the os.seek and seek_end are compatible with python 2 and
3.

Deb

On Fri, Jul 15, 2016 at 12:10 AM, Noah van Dresser <notifications@github.com

PR #453, opened 2016-07-12
Multiple .ini files can be used now.

### LINKED ISSUE #436: client accepts multiple .ini files but only runs the last .ini file
Description of problem: 

MTT client accepts multiple .ini files but only runs the last .ini file

Version-Release number of selected component (if applicable): 

MTT commit 84b227219f78125fab10a56d11d59bbe92dced28

Steps to Reproduce:
 /home/test/jenkins/mtt/pyclient/pymtt.py --base-dir=/home/test/jenkins/mtt/pylib <inifile#1>.ini <inifile#2>.ini

Actual results:

 Only the final .ini file is run.

Expected results: 
 
 All .ini files should be run

Additional info:

DebRez:
fixes #436

### DIFF
--- a/pylib/System/TestDef.py
+++ b/pylib/System/TestDef.py
@@ -437,21 +437,20 @@ class TestDef(object):
             if not os.path.isfile(testFile):
                 print("Test .ini file not found!: " + testFile)
                 sys.exit(1)
-##TO DO, remove from loop, change testFile to list of files               
-            self.config = configparser.ConfigParser()
-            self.config.read(testFile)
-            for section in self.config.sections():
-                if self.logger is not None:
-                    self.logger.verbose_print("SECTION: " + section)
-                    self.logger.verbose_print(self.config.items(section))
-                if self.options['dryrun']:
-                    continue
-                if section.startswith("SKIP") or section.startswith("skip"):
-                    # users often want to temporarily ignore a section
-                    # of their test definition file, but don't want to
-                    # remove it lest they forget what it did. So let
-                    # them just mark the section as "skip" to be ignored
-                    continue;
+        self.config = configparser.ConfigParser()
+        self.config.read(self.log['inifiles'])
+        for section in self.config.sections():
+            if self.logger is not None:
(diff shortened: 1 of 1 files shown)


DebRez:
@rhc54 
@noahv

pylib/System/TestDef.py
```diff
@@ -437,21 +437,20 @@ def configTest(self):
...
-                if self.options['dryrun']:
-                    continue
-                if section.startswith("SKIP") or section.startswith("skip"):
-                    # users often want to temporarily ignore a section
-                    # of their test definition file, but don't want to
-                    # remove it lest they forget what it did. So let
-                    # them just mark the section as "skip" to be ignored
-                    continue;
+        self.config = configparser.ConfigParser()
+        self.config.read(self.log['inifiles'])
```
rhc54: I'm a little confused as to how this solves the problem. In line 435, we set ```self.log['inifiles']``` to only  ```args.ini_files[0]```. Is that value the first file in the list? Or is it the entire list? If the former, then how does this result in reading all the .ini files?

DebRez: self.config.read(testFile) was part of the loop checking to see if the .ini files existed.  I took the self.config.read out of the loop and used the list of files from the new argparse args.ini_files which was producing a list of lists so I needed to add the [0] to get to the actual list of files. ConfigParser will take either a single file or a list of files but it will not take a list of lists.

DebRez: I put a comment in the TestDef.py change file,  I am not sure if you get an
email when that happens or not.  Here are the contents of the comment:

self.config.read(testFile) was part of the loop checking to see if the .ini
files existed. I took the self.config.read out of the loop and used the
list of files from the new argparse args.ini_files which was producing a
list of lists so I needed to add the [0] to get to the actual list of
files. ConfigParser will take either a single file or a list of files but
it will not take a list of lists.

Deb Rezanka

rhc54: Ah, okay - thanks for the explanation!

Issue #444, opened 2016-06-28
Cherrypy expects wrong path for virtualenv site-packages


ribab:
The cherrypy server installation script located in `mtt/server/php/cherrypy/bin/mtt_server_install.py` expects to find the `site-packages` folder it generated using virtualenv within the `src/env/lib/python/site-packages/` path, however the path that is actually generated by virtualenv is `src/env/lib/python2.7/site-packages/`.

A simple fix is to change the text "python" to "python2.7" in this line in the "mtt_server_install.py" file mentioned above:

    call(['cp', '-rs', cwd + '/' + source_dir + '/' + package_name,
      env_dir + '/lib/python/site-packages/' + package_name])

I am hesitant for this fix since I am unsure whether python virtualenv generates `src/env/lib/python2.7/site-packages/` for everyone, or if some systems use the `src/env/lib/python/site-packages/` path for their version of virtualenv.

jjhursey:
Yeah, we should investigate this some more. I suspect that it's environment specific (so will match the version of python being used). We might be able to make the script smarter to detect the name of the environment and use it in the path construction.

jjhursey:
It looks like virtualenv has a `cdsitepackages` key that might solve this:
  * http://docs.python-guide.org/en/latest/dev/virtualenvs/

Issue #443, opened 2016-06-22
IUDatabase Reporter plugin missing fields


jjhursey:
To submit to the IU MTT Database (via the REST interface) the python client needs to submit all of the required fields. The REST interface will reject the submission if a required field is missing.

In the file:
  https://github.com/open-mpi/mtt/blob/master/pylib/Stages/Reporter/IUDatabase.py

Near lines 186, 304, 391 (search for `JJH Todo`) you will see the `data` packet being constructed to send to the server. In the list that follows all of the `data['FIELD'] = None` (where `FIELD` is something like `compiler_name`) lines (that are not commented out) need to be filled in.

For example (lines 212-226):
```python
            data['full_command'] = None

            # For now just mark the time when submitted
            data['start_timestamp'] = datetime.utcnow().strftime("%c")

            data['result_message'] = None
            data['test_result'] = None

            try:
                data['exit_value'] = lg['status']
            except KeyError:
                data['exit_value'] = None

            # Optional
            # data['duration'] = None
```
The fields that need to be filled in are
 - `full_command`
 - `start_timestamp` (it's already filled in here)
 - `result_message`
 - `test_result`
 - `exit_value` (already filled in here)

You can optionally set `duration`, but it is not necessary to correctly submit to the database.

If the submit was successful then you will get a successful message from the server, and will be able to see the results in the MTT Reporter/Viewer website.

Some of these fields need to be documented better (feel free to ask questions here or on the mailing list about fields that need clarification), but the `test_result` takes an integer value:
 * 0 : Failed
 * 1 : Pass
 * 2 : Skipped
 * 3 : Timed out

Documentation on the Submit REST interface is here:
  https://github.com/jjhursey/mtt-modern/wiki/REST-Interface-(Submit-Side)

PR #456, opened 2016-07-18
Fixed IUDatabase plugin missing fields (Issue #443)

ribab:
Also fixed cherrypy path (Issue #444)

Also fixed JunitXML byte encoding issue by encoding XML output to utf-8 format


pylib/Tools/Build/Autotools.py
```diff
@@ -197,6 +199,16 @@ def execute(self, log, keyvals, testDef):
                     # this is a multistep operation, and so we need to
                     # retain the output from each step in the log
                     log['autogen'] = (stdout, stderr)
+
+                    # Use mpi4py to get MPI information for IUDatabase plugin
+                    log['mpi_name'] = MPI.get_vendor()[0]
```
rhc54: Given that mpi4py is open source, can we look at its license and (if compatible) simply extract the content of this function? It would be preferable to including the entire mpi4py library just to call this one thing.

ribab:
mpi4py get_vendor() function simply calls a C function using something like cython. The C function then makes use of macros defined at compile time that tells it which MPI version is running.

My best solution using pure python would be to parse the output of "mpirun --version", which isn't as clean of a solution as what mpi4py does since "mpirun --version" doesn't have a standard output.

pylib/Tools/Build/Autotools.py
ribab: mpi4py get_vendor() function simply calls a C function using something like cython. The C function then makes use of macros defined at compile time that tells it which MPI version is running.

My best solution using pure python would be to parse the output of "mpirun --version", which isn't as clean of a solution as what mpi4py does since "mpirun --version" doesn't have a standard output.

rhc54: IIRC, mpi4py has to be configured and built against an MPI, yes? And so the macros are setup at time of compiling mpi4py, which means they will only tell you information about the MPI which mpi4py was built against - it doesn't tell you the version of MPI you have in your path at the moment.

Is that not true? I don't think you can just "point" mpi4py to any MPI at runtime and have it work.

ribab: That is true. If the MPI version is swapped out or re-installed, then mpi4py would have to be re-installed.

I will work on an implementation using "mpirun --version" and show it to Noah when he's available.

rhc54:
Probably worth doing a ```git rebase -i master``` on this to squash it down to one commit - you can leave all the commit messages so we see what was done

This looks fine to me. At some point, we might want to optimize by not checking for MPI version on every invocation of the Autotools and Shell plugins as that seems wasteful. We only need that info for particular stages, and so it might be more efficient to gather it for that specific stage. Still, that's an optimization.

noahv:
Looks great.  Yes, please squash down to one commit and ship it.  Awesome job Ricky!  

@jjhursey please note that this includes a server side change that makes the MPI install table "configure_arguments" optional instead of required.  Should we update this page?

https://github.com/jjhursey/mtt-modern/wiki/REST-Interface-(Submit-Side)

That enables usage of the default mpi install from the system under test, environment module loaded mpi, or middleware built mpi install.

jjhursey:
Wait - why are `configure_args` not passed? That's a pretty important aspect of the build. Is it jus tin certain situations or all the time?

You could pass a sentinel value for an already installed configuration ("Already Installed" or such) if you wanted. But for a build that includes `configure` we need to see those values.

noahv:
The intent was that if a middleware build was included that the configure_args would also be submitted.  Let me double check.

rhc54:
If we are trying to deal with an "as-is" build, remember that you can retrieve the configure args by just running ```head config.log``` and looking for the 7th line.

noahv:
Unfortunately the configure_arguments is not being populated even when specified as part of the middleware build.

Ricky will file an issue and get this resolved.

Here is the fix:

--- a/pylib/Stages/Reporter/IUDatabase.py
+++ b/pylib/Stages/Reporter/IUDatabase.py
@@ -502,7 +502,7 @@ class IUDatabase(ReporterMTTStage):
                 data['mpi_version'] = None
 
         try:
-            data['configure_arguments'] = lg['configure_options']
+            data['configure_arguments'] = logger.getLog(lg['middleware'])['configure_options'] 
         except KeyError:
             data['configure_arguments'] = None

PR #458, opened 2016-07-19
Add clck_1 view for Intel(R) Cluster Checker

atulkulk:
I have added a new db view(clck_1) to work with Intel(R) Cluster Checker tool.

### DIFF
--- a/server/sql/analysis/clck_view.sql
+++ b/server/sql/analysis/clck_view.sql
@@ -1,0 +1,22 @@
+--
+-- View for Intel(R) Cluster Checker
+--
+DROP VIEW IF EXISTS clck_1;
+CREATE OR REPLACE VIEW clck_1 AS
+    SELECT test_run.duration,
+        0 AS encoding,
+        test_run.exit_value AS exit_status,
+        submit.hostname,
+        ''::text AS node_names,
+        0 AS num_nodes,
+        test_run.test_run_command_id::character varying AS optionid,
+        test_run.test_run_id AS rowid,
+        test_names.test_name AS provider,
+        test_run.result_stdout AS stdout,
+        test_run.result_stderr AS stderr,
+        date_part('epoch'::text, test_run.start_timestamp)::integer AS "timestamp",
+        submit.local_username AS username,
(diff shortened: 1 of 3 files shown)


rhc54:
@jjhursey Does this look okay to you?

hppritcha:
Is this for the project in general or specific to v4.0?

rhc54:
it pertains to the MTT project in general - will support either Perl or Python clients

server/sql/analysis/clck_view.sql
```diff
@@ -0,0 +1,22 @@
...
+DROP VIEW IF EXISTS clck_1;
+CREATE OR REPLACE VIEW clck_1 AS
+    SELECT test_run.duration,
+        0 AS encoding,
+        test_run.exit_value AS exit_status,
+        submit.hostname,
+        ''::text AS node_names,
+        0 AS num_nodes,
+        test_run.test_run_command_id::character varying AS optionid,
+        test_run.ctid AS rowid,
```
jjhursey: That is `test_run.ctid` I don't see that in the schema:
 * https://github.com/open-mpi/mtt/blob/master/server/sql/schemas-v3.sql

jjhursey:
Please rebase this into 1 commit before merging.
I had one comment about a field that I don't think exists in the schema. Maybe it was a local addition that didn't make it to the repo yet?

server/sql/analysis/clck_view.sql
atulkulk: ctid is a system column we used as a unique row identifier to test with CLCK. You can find the definition of ctid at [this](https://www.postgresql.org/docs/9.5/static/ddl-system-columns.html) link. However, after discussing with the team, we will use the existing test_run_id column as a unique row identifier. I will update the view.

jjhursey: Yeah, since we are being pushed to move away from posgresql, I would prefer to not use a system constant here. `test_run_id` should be sufficient.

jjhursey:
This looks good to me. 👍

PR #462, opened 2016-08-04
Fixed issues 459, 460, 461

ribab:
The MPIVersion plugin can now be used as a regular user, and the configure_arguments field is now populated.

Fixed issue 459: get_mpi_version is auto-generated so that the user running MTT has access to the file

Fixed issue 460: MPIVersion marks the name and version of MPI as "None" and "Unknown" if mpicc is not found

Fixed issue 461: configure_arguments is now populated in the database submitter


rhc54:
👍 

Thanks!

Issue #464, opened 2016-08-11
OpenMPI.py plugin missing log entries


ribab:
The OpenMPI.py plugin does not have an entry for "np" (or number-of-processes) into the log. This plugin needs to be compared with the SLURM plugin to find out what it is missing.

PR #465, opened 2016-08-22
Added np to log

DebRez:
OpenMPI.py now has an entry for "np"

This fixes issue #464 

@ribab 
@rhc54 
@hppritcha


rhc54:
I think the logic looks fine, but the indentation is wrong?

DebRez:
I use emacs and it looks fine there but I opened it in vim and was able to
see and fix the problem.  I will check my indentation rules in emacs and
see if I set something wrong but this hasn't happened before.  Any ideas
what may cause emacs and vim to show spacing differently?

DebRez:
fixed spacing

rhc54:
Check your tabs settings. We ask that everyone set their editors to replace tabs with spaces, and set the mapping to 4 spaces/tab stop. It's likely one of those that's the issue.

PR #466, opened 2016-08-26
Cleanup several bugs that prevented the Python client from executing cleanly:

rhc54:
* do not require a .ini file just to ask for --list-stages and other info cmds

* move the compiler output line to debug prints

* properly parse the OMPI cmd line into a list

* properly call MPI_Finalize at the end of mpi_get_version so we don't segfault


rhc54:
@DebRez @ribab  Would you folks mind taking a quick peek?

rhc54:
I'm going to bring this in so I can continue resolving the OMPI submission problems

PR #467, opened 2016-08-27
Continue push towards provisioning support

rhc54:
Rename the Warewulf provisioning plugin to WWulf3 to distinguish it from upcoming versions. Add a new Command-N-Control category of tools, and implement a plugin that uses ipmitool to execute IPMI commands such as "power cycle"

@noahv


noahv:
Merged.  Thanks.  -- Noah

PR #470, opened 2016-09-02
Fix the multi-test.ini file option so we correctly ingest multiple files

rhc54:
Fix the multi-plugin directory path option, and provide a new option that lets you ignore plugin directory errors in case you want to specify a bunch of directories, even if some don't exist on a given system

Enable the --sections and --skip-sections options

@noahv


noahv:
Looks good.  Thanks,  -- Noah

PR #471, opened 2016-09-06
Prototype the Warewulf 3 support. 

rhc54:
Add a function to TestDef to select a plugin given a name and category - will eventually propagate use of it to remove redundant code. Update IPMITool to handle need for "sudo" to execute commands

@noahv See what you think


noahv:
Looks good; thanks for adding the selectPlugin function to avoid duplicating code.

Feel free to merge this when you are ready.

noahv:
Can we merge this now?

rhc54:
sure - you'll need to test it in operation, though, as i haven't been able to do so (can't take the chance on knocking my cluster offline)

PR #473, opened 2016-09-07
Added --stop-on-fail. If a stage fails, exit and issue a non-zero ret…

noahv:
…urn code.


rhc54:
👍

PR #474, opened 2016-09-09
Do a better job of parsing command line options when converting to match the intended data type

rhc54:
@noahv @DebRez Please take a look at this - we were overwriting options with strings regardless of the intended type, and so comparisons against int values were failing


DebRez:
This looks good to me.

PR #478, opened 2016-09-30
Added markup for main doxygen page and doxygen grouping of stages, to…

noahv:
…ols, and utilities.

Documents plugin options as doxygen parameters.


rhc54:
Very nice!

PR #480, opened 2016-10-12
Corrected , to + in print statement in Logger

DebRez:
was causing an ascii encode error when using --verbose

@noahv


DebRez:
oops, printing to file, still looking for the source of my ascii error

PR #481, opened 2016-10-24
Add Test Run "corefile_size_limit" param to the perl MTT client

jsquyres:
If set to >= 0, setrlimit to limit corefiles to that size.  If set to a negative number, corefile sizes are unlimited.  The default is unlimited corefiles.

This functionality requires the `BSD::Resource` perl module.  If you don't have it (and we therefore can't set corefile size limits), you'll get an error (because it would be worse to generate gigabytes / terrabytes of corefiles when you specifically asked us not to).

@jjhursey Please review.

### DIFF
--- a/lib/MTT/Test/Run.pm
+++ b/lib/MTT/Test/Run.pm
@@ -525,6 +525,12 @@ sub _do_run {
     $tmp = $ini->val($section, "treat_timeouts_as_fail");
     $config->{treat_timeouts_as_fail} = $tmp
         if (defined($tmp));
+    $config->{corefile_size_limit} = -1;
+    $tmp = $ini->val($section, "corefile_size_limit");
+    $config->{corefile_size_limit} = $tmp
+        if (defined($tmp));
+
+    MTT::DoCommand::set_corefile_size_limit($config->{corefile_size_limit});
 
     $MTT::Test::Run::mpi_details = undef;
     foreach my $field ($ini->Parameters($section)) {
--- a/lib/MTT/DoCommand.pm
+++ b/lib/MTT/DoCommand.pm
@@ -33,6 +33,21 @@ use File::Spec;
 use Cwd;
 use Benchmark;
 
+# See if we've got BSD::Resource.  If so, get some values from it so
+# that we don't get compile errors for systems that *don't* have it.
+my $_can_setrlimit;
+my $_rlimit_core;
+my $_rlim_infinity;
+if (eval "require BSD::Resource") {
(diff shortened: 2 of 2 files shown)


rhc54:
tsk...tsk...you should have done this as a Python plugin, dude!

jsquyres:
@rhc54 ...if it worked 😄   You guys haven't told me that it's viable to switch yet.

lib/MTT/Test/Run.pm
```diff
@@ -525,6 +525,12 @@ sub _do_run {
     $tmp = $ini->val($section, "treat_timeouts_as_fail");
     $config->{treat_timeouts_as_fail} = $tmp
         if (defined($tmp));
+    $config->{corefile_size_limit} = -1;
+    $tmp = $ini->val($section, "corefile_size_limit");
+    $config->{corefile_size_limit} = $tmp
+        if (defined($tmp));
+
+    MTT::DoCommand::set_corefile_size_limit($config->{corefile_size_limit});
```
jjhursey: I'm reading the [perl manual page](http://search.cpan.org/~jhi/BSD-Resource-1.2910/Resource.pm#ERRORS) for `BSD::Resource` and the notes under **Errors** and **Known Issues** are a bit concerning with environment support for `setrlimit`.

In the case where the INI file does not specify `corefile_size_limit` then we should not be calling `MTT::DoCommand::set_corefile_size_limit`. This allows folks with unsupported environments to still run MTT, but without this feature.

I suggest changing this line to:
```perl
MTT::DoCommand::set_corefile_size_limit($config->{corefile_size_limit})
   if (defined($tmp));
```

or, slightly better,
```perl
if (defined($tmp)) {
    $config->{corefile_size_limit} = $tmp;
    MTT::DoCommand::set_corefile_size_limit($config->{corefile_size_limit});
}
```

jsquyres: Sure, that sounds reasonable.

jsquyres: Actually, I forgot: there's a reason I set it this way.

If you set a max corefile value for one Test Run, but *don't* set it for another, we want to run it again in the section where you *don't* set it so that it gets set back to an infinite core file size.

I didn't actually check `setrlimit` for an error, so it should be ok even if the current environment doesn't support it.

jsquyres:
I did just push one bug fix that will still allow the default "infinite" setting to work properly even in environments that don't have `BSD::Resource`.

PR #494, opened 2017-01-04
Add an ALPS launcher plugin

rhc54:
Pretty much a carbon copy of SLURM, with a few cmd line option changes

### DIFF
--- a/pylib/Tools/Launcher/ALPS.py
+++ b/pylib/Tools/Launcher/ALPS.py
@@ -15,12 +15,11 @@ from LauncherMTTTool import *
 ## @addtogroup Tools
 # @{
 # @addtogroup Launcher
-# @section SLURM
+# @section ALPS
 # @param merge_stdout_stderr       Merge stdout and stderr into one output stream
 # @param skipped                   Exit status of a test that declares it was skipped
 # @param hostfile                  The hostfile for OpenMPI to use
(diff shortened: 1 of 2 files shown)


hppritcha:
Thanks for doing this.  It will be difficult for me to test any time soon because I've not got my python client setup yet on any cray machine using aprun.

ribab:
Hi Ralph, we have this running, but it needs this patch so we can specify multiple options to aprun. The SLURM plugin has the same issue and I will fix that one.
```patch
diff --git a/pylib/Tools/Launcher/ALPS.py b/pylib/Tools/Launcher/ALPS.py
index 4a08e28..c303a3b 100755
--- a/pylib/Tools/Launcher/ALPS.py
+++ b/pylib/Tools/Launcher/ALPS.py
@@ -279,7 +279,8 @@ class ALPS(LauncherMTTTool):
         # assemble the command
         cmdargs = [cmds['command']]
         if cmds['options'] is not None:
-            cmdargs.append(cmds['options'])
+            for op in cmds['options'].split():
+                cmdargs.append(op)
         if cmds['np'] is not None:
             cmdargs.append("-n")
             cmdargs.append(cmds['np'])
```

rhc54:
@ribab Thanks! I will modify and commit

Issue #468, opened 2016-08-31
result_stderr not packaged properly in Pyclient for submit


jjhursey:
See the conversation here:
- https://mail-archive.com/mtt-devel@lists.open-mpi.org/msg00686.html

It looks like the `result_stderr` key is passed as an array containing one string, instead of passing just the string.

Note that `result_stdout` is being packaged properly.

ribab:
The problem is that the "join" command is used in IUDatabase.py, and previous steps already converted stderr to a string.

I will look through the rest of the code and make sure that both stdout and stderr are handled consistently.

rhc54:
Just FYI: the execute_cmd plugin now returns stdout/stderr as lists. This was done in order to support the truncation directive. I adjusted the code in IUDatabase to properly handle stdout/stderr submission, but code that uses execute_cmd will need similar adjustments

hppritcha:
@DebRez if you have time could you take a look at this?

PR #496, opened 2017-01-26
Fixed fields passed into reporter

ribab:
Reporter wasn't working since some fields were passed in as lists. I fixed this by joining the lists together.

I also realized that the "result_stderr" bad-format problem wasn't occurring. Maybe that bug was already fixed. -- Bug #468


ribab:
It says that my commit wasn't signed off. I'll fix that later

jjhursey:
So you are able to submit to the MTT server just fine now?

ribab:
I didn't test against an official MTT server. I tested against locally hosted DB and Cherrypy server.

ribab:
It should be good to merge now. I fixed the signing error.

PR #498, opened 2017-02-01
Add harasser utility

rhc54:
I haven't been able to figure out how to deal with the Python 2.7 vs Python 3.4 compatibility issue for one key place. In testDef.py, line 337, we need to set the multiprocessing launch method to "spawn" so we ensure that the subprocess is in its own process. This ensures that the Python global lock doesn't take affect and prevent the new process from making independent progress.

Could use some help in that regard.


noahv:
Alternate approach already in place for the harasser thread support.

PR #500, opened 2017-02-03
Fixed db reporter in a way that doesn't break text-file reporter

ribab:
Previous fix for DB reporter caused the text-file reporter to not work and output only 1 character per line for stdout.

I undid the previous fix changes outside of IUDatabase.py and fixed the reporter fields within only the IUDatabase.py file. This way, the text-file reporter is unaffected by my fix and the output works again.

Previous pull request which broke text-file reporter: https://github.com/open-mpi/mtt/pull/496


noahv:
Looks good; thanks!  -- Noah

PR #512, opened 2017-04-10
Execute module unload first then module load.

### LINKED ISSUE #511: The module and module_unload commands (supported in Shell, ALPS, and SLURM) execute in the wrong order
The module and module_unload commands (supported in Shell, ALPS, and SLURM) execute in the wrong order.

This causes the attempt to load to fail because you can not have more than one mpi environment module loaded.

noahv:
Fixes #511.

### DIFF
--- a/pylib/Tools/Build/Shell.py
+++ b/pylib/Tools/Build/Shell.py
@@ -179,18 +179,6 @@ class Shell(BuildMTTTool):
 
         # check to see if they specified a module to use
         # where the compiler can be found
-        usedModule = False
-        try:
-            if cmds['modules'] is not None:
-                status,stdout,stderr = testDef.modcmd.loadModules(cmds['modules'], testDef)
-                if 0 != status:
-                    log['status'] = status
-                    log['stderr'] = stderr
-                    return
-                usedModule = True
-        except KeyError:
-            # not required to provide a module
-            pass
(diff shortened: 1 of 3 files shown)


rhc54:
Just for my own clarity: this basically requires that someone list the modules to unload and to load for a given stage, yes? If true, I suppose it would be nice, but probably too difficult, to detect that a new module is going to conflict with one that was previously loaded and automatically do the unload.

That said, it strikes me that I had tried to always unload any module that was loaded for a given stage. So is this change required for dealing with modules that were loaded before the MTT run was started? I guess we didn't provide a way of "unloading" things, did we - oops 🤦‍♂️

noahv:
For the plugins I mentioned (ALPS, SLURM, Shell) I did add support for requesting a specific module be unloaded in a previous patch.  

That support also includes re-loading that module once the stage completes.

There are a few MTT plugins which don't have the unload support (Autotools, AlreadyInstalled, Git). 
 
How about we add an enhancement that adds module_unload support to those modules and in that same enhancement request we can also add the requirement to automatically check for and resolve conflicts?

I agree that we would want to automatically detect conflicts to make the user experience better.  That said, can we push this patch?

rhc54:
Oh sure - please go ahead and commit. I was just thinking aloud about the overall issue you were addressing.

Issue #513, opened 2017-04-10
Compiler version sometimes doesn't fit inside database


ribab:
In "Compilers.py", the compiler version logged is simply the entire output of typing `cc --version` (or whatever equivalent of `cc` for version of compiler used).

The issue is sometimes the output of the above command was longer than 64 bytes and doesn't fit into the database. For example, the output of intel compiler `icc --version` is this:

$ icc --version
icc (ICC) 17.0.1 20161005
Copyright (C) 1985-2016 Intel Corporation.  All rights reserved.

$ icc --version | wc -c
92

which is larger than 64 bytes.

PR #514, opened 2017-04-10
Fixes Issue #513: Compiler version truncated to fit in database variable

ribab:
The compiler version was sometimes longer than 64 bytes (depending on the version, sometimes copyright information and other types of information were also included).
Truncating the compiler version to be max 64 bytes allows it to always fit within the database while still capturing enough information to determine compiler version.

Another thing that was fixed was that exit_value was sometimes not logged to DB. I fixed this by adding some more conditionals to ensure that exit_value is always logged.

### DIFF
--- a/pylib/Stages/Reporter/IUDatabase.py
+++ b/pylib/Stages/Reporter/IUDatabase.py
@@ -288,6 +288,8 @@ class IUDatabase(ReporterMTTStage):
                             data['exit_value'] = int(lgerr.split("[Errno ")[1].split("]")[0])
                         except:
                             data['exit_value'] = -1
+                    else:
+                        data['exit_value'] = -1
                 else:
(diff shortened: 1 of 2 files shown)


rhc54:
Don't you just want to grab the first line of the output, and then constrain that to 64 bytes?

ribab:
I wasn't sure if the version is always in the first line. If that's the case then that would work great

rhc54:
It is always on the first line for the compilers I quickly checked:
```shell
$ icc --version
icc (ICC) 14.0.3 20140415
$ gcc --version
gcc (GCC) 6.1.0
```
I think I'd go with that until someone proves otherwise - @jjhursey do the xlc compilers adhere to that practice?

ribab:
I changed it to grab only the first line.

PR #518, opened 2017-04-25
Any delimiter (not just ":") can now be used in INI files

ribab:
Code before required ":" to be used as a delimiter between stage and name. Based on how the information was stored in the log, the delimiter didn't make sense. Also, the title (title=stage+name) with ":" in it was causing an issue with some makefiles. This fix provides freedom in title naming.

[MTTDefaultsIsAVeryCoolStage] is now a valid title.

[TestRun--projectAwesome] is also now a valid title.

As well as [TestGet:GrabMyFiles]

### DIFF
--- a/pylib/Stages/Reporter/IUDatabase.py
+++ b/pylib/Stages/Reporter/IUDatabase.py
@@ -128,7 +128,8 @@ class IUDatabase(ReporterMTTStage):
 
         data = {}
 
-        profile = testDef.logger.getLog('Profile:Installed')
+        profile = [lg for lg in testDef.logger.getLog(None) if 'Profile' in lg['section'] and 'Installed' in lg['section']][0]
+#        profile = testDef.logger.getLog('Profile:Installed')
         metadata = {}
         metadata['client_serial'] = client_serial
         metadata['hostname'] = "\n".join(profile['profile']['nodeName'])
--- a/pylib/Stages/Reporter/IUDatabase.py
+++ b/pylib/Stages/Reporter/IUDatabase.py
@@ -129,7 +129,6 @@ class IUDatabase(ReporterMTTStage):
         data = {}
 
         profile = [lg for lg in testDef.logger.getLog(None) if 'Profile' in lg['section'] and 'Installed' in lg['section']][0]
-#        profile = testDef.logger.getLog('Profile:Installed')
         metadata = {}
         metadata['client_serial'] = client_serial
         metadata['hostname'] = "\n".join(profile['profile']['nodeName'])
@@ -212,8 +211,6 @@ class IUDatabase(ReporterMTTStage):
         # For now assume that we only had one test_build submitted
         # and all of the tests that follow are from that test_build
         common_data['test_build_id'] = test_info['test_build_id']
-
-        #import pdb; pdb.set_trace()
 
         for trun in (lg['testresults'] if 'testresults' in lg else [lg]):
             data = {}
@@ -517,7 +514,7 @@ class IUDatabase(ReporterMTTStage):
         options = None
 
         # get the system profile
-        profile = logger.getLog('Profile:Installed')['profile']
+        profile = [lg for lg in testDef.logger.getLog(None) if 'Profile' in lg['section'] and 'Installed' in lg['section']][0]
         if profile is None:
             print("Error: Failed to get 'profile'")
             return None
--- a/pylib/Tools/Executor/sequential.py
+++ b/pylib/Tools/Executor/sequential.py
@@ -80,12 +80,17 @@ class SequentialEx(ExecutorMTTTool):
                 # this section
                 if "SKIP" in title:
                     continue
+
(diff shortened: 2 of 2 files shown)


rhc54:
I'm scratching my head over this one. The separator was there so we could extract the name of the stage from the title, thus giving us a way to find the correct plugin for that stage. How does this happen now?

ribab:
Even before my change, the order at which the sections are ran has nothing to do with the colon. The sequential executor simply checks if the current stage name in the stage order exists within the section title; if it exists, the stage is executed.

I parse out the stage name by using the stage name as a delimiter instead of using the colon, and then rebuilding the stage name while keeping the "ASIS" keyword if it exists (since "ASIS" is checked for later).

((In the code, variable "step" is the stage name and variable "title" is the section title))

ribab:
Deleted last comment since I misunderstood your question

* The name of the stage is given in "testDef.loader.stageOrder" (called "step" in the code)
* Stage is parsed out using the stage name given as the delimiter rather than using colon as the delimiter

rhc54:
So you look for the start of the title to match a stage name, yes? It should be okay.

FWIW: I don't know if the code changed, but the way it used to work was that we executed the stages in the priority ordering given in testDef when it loads the configuration. I forget if it went from low-to-high or the other way around, but we read all the stages in and then executed them in stage order. Within a given stage, we would execute them in the order they were given.

So if you had multiple ```StageFoo:xxx``` entries, where xxx was the title of the specific StageFoo entry, then we would execute those in the order of appearance in the .ini file. However, all the StageFoo entries would execute after StageBar, regardless of ordering in the file, if StageBar was configured to have precedence.

HTH  
Ralph

rhc54:
Actually, that may be the combinatorial executor I described - I seem to recall making a design decision that the sequential executor just did things in order of appearance.

jsquyres:
Just curious: what's the need for this?  I don't object, but it does seem to be spurious.  I.e., what's the need for `[MTTDefaultsIsAVeryCoolStage]` to be a valid title?  Doing a regular expression based on `:` was simple, easy to read, and guaranteed easy-to-read INI files.

Additionally, please squash your two commits together before merging to master.  It's a little weird to commit a mistake, then add a commit to fix the mistake, and then commit both of those (including the initial mistake) to master when it could have been fixed on the PR ahead of time.

Thank you!

rhc54:
Apparently, someone had a problem in their environment with the colon. Surprising to me, and perhaps a little more detail as to why it occurred would be helpful. Personally, I prefer having a delimiter as I agree it makes it easier to read.

noahv:
Yeah, when trying to build [https://github.com/linux-test-project/ltp](url) tests which use Autotools we ran into trouble with the ":" in the path.  Autotools can be challenging to debug so we went with this instead.

```
root@pluto1:/home/test/testit/TestGet:LTP/ltp-full-20170116 # make
/home/test/testit/TestGet:LTP/ltp-full-20170116/include/mk/automake.mk:108: *** target pattern contains no '%'.  Stop.
root@pluto1:/home/test/testit/TestGet:LTP/ltp-full-20170116 # 
```
Here is that line in the automake.mk
```
$(AUTOGENERATED_FILES): $(top_builddir)/config.status
        $(SHELL) $^
```
We can dig further into this if needed ...

jsquyres:
Ah, I see.

In the perl client, I'm pretty sure we `s/:/_/g` in the filename version of the section names to avoid that kind of path problem.

FWIW, what if my INI file still uses a `:` -- is that still going to be a problem in some environments?  Because if so, you should either do the `s/:/_/g` trick, or you should prohibit using `:` altogether (keeping in mind that this introduces an incompatibility with the Perl client -- which is not the end of the world, but it is another thing that Perl client users [like me!] will need to do to someday switch over to the python client).

rhc54:
I'm puzzled - why would this create a conflict? The autotools should only be parsing the test suite, not the Python plugins or main code, nor the .ini files. Is the problem more likely that there is an error in the code trying to build those tests?

For example, I see that your path includes the stage title, including the colon. So if we just do the sub @jsquyres suggests, then the problem goes away, yes? I would definitely prefer that solution.

jsquyres:
@rhc54 I think it's not actually an Autotools conflict -- it's a combination of two things:

1. The Python client made a directory with a `:` in it.
1. That directory was used as part of a larger filename to the left hand side of a Makefile `target: dependencies` line, hence the `:` in the target specification filename became a problem (i.e., there were two `:`'s on the line, and `make` went "Huh?")

I think the real solution is to ensure that the Python client never makes a directory with a `:` in it.

ribab:
I'm working on making ":" = "_" in path. It will be a new pull request.

ribab:
Hi @jsquyres , I made ":" = "_" in path. Take a look at pull request https://github.com/open-mpi/mtt/pull/527

jsquyres:
@ribab Cool.  Does that make this PR moot?

ribab:
@jsquyres Yea this PR is moot now. I'm going to close.

PR #519, opened 2017-04-26
Test list implemented for TestRun plugins

ribab:
SLURM.py, OpenMPI.py, and ALPS.py now have a "test_list" option which specifies a list of tests to run. The list is of filenames (no paths).

If test_list is not specified, all tests will be run.

Also, an error in ALPS plugin was fixed regarding using "np" option rather than using "options" option.


pylib/Tools/Launcher/ALPS.py
```diff
@@ -224,6 +226,13 @@ def execute(self, log, keyvals, testDef):
         # parse any provided options - these will override the defaults
         cmds = {}
         testDef.parseOptions(log, self.options, keyvals, cmds)
+
+        # Check if command is correct
+        if cmds['command'] != "aprun":
+            log['status'] = 1
+            log['stderr'] = "Command for SLURM plugin must be aprun. It is currently '%s'" % (cmds['command'] if cmds['command'] is not None else "None")
```
rhc54: I think you have a copy/paste error here - shouldn't that be "ALPS plugin"?

PR #521, opened 2017-04-26
Add tables to DB for BIOS, FW, Provisioning, and Harassers

ribab:
WARNING: I am unsure if this pull request will break the Perl client since it adds to the MTT DB SQL schema.

This pull request is a good example of the effort it takes to make database changes while keeping compatibility with the python client.

Tables are added for the BIOS, FW, and Provisioning stages, along with Harasser table, and are connected to the Test Run table, since each of these only affect the test while it is running.
* BIOS, FW, Provisioning, and Harassers affect the compute nodes which are only accessed when the test is running

There is no Harasser stage, and Harasser framework is not developed yet, however I included it in the DB anyways.

### DIFF
--- a/pylib/Stages/Provisioning/WWulf3.py
+++ b/pylib/Stages/Provisioning/WWulf3.py
@@ -67,6 +67,18 @@ class WWulf3(ProvisionMTTStage):
         # parse what we were given against our defined options
         cmds = {}
         testDef.parseOptions(log, self.options, keyvals, cmds)
+
+        mylog = {}
+        if cmds['target']:
+            mylog['target'] = cmds['target']
+        if cmds['image']:
+            mylog['image'] = cmds['image']
+        if cmds['controller']:
+            mylog['controller'] = cmds['controller']
+        if cmds['bootstrap']:
+            mylog['bootstrap'] = cmds['bootstrap']
+        log['provisioning'] = mylog
+
         # they had to at least give us one target node and controller
         try:
             if cmds['target'] is None:
--- a/pylib/Stages/Reporter/IUDatabase.py
+++ b/pylib/Stages/Reporter/IUDatabase.py
@@ -191,6 +191,7 @@ class IUDatabase(ReporterMTTStage):
                                       logger.getLog(self._extract_param(logger, lg['section'], 'parent')),
                                       metadata,
                                       s, url, httpauth)
+
         if test_info is None:
             return None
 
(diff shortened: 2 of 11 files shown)


jsquyres:
Please do not merge until you know it will not break the Perl client on the community database.

There's still a large number of people using the Perl client (because we've never been told that the Python client is ready), so maintaining compatibility with the current production database is unfortunately necessary.

pylib/Stages/Provisioning/WWulf3.py
```diff
@@ -67,6 +67,14 @@ def execute(self, log, keyvals, testDef):
         # parse what we were given against our defined options
         cmds = {}
         testDef.parseOptions(log, self.options, keyvals, cmds)
+
+        mylog = {}
+        if cmds['target']: mylog['target'] = cmds['target']
+        if cmds['image']: mylog['image'] = cmds['image']
+        if cmds['controller']: mylog['controller'] = cmds['controller']
+        if cmds['bootstrap']: mylog['bootstrap'] = cmds['bootstrap']
```
rhc54: Perhaps for consistency you should split the above if statements into two lines like we do everywhere else?

pylib/System/TestDef.py
```diff
@@ -218,6 +218,7 @@ def parseOptions(self, log, options, keyvals, target):
             log['status'] = 1
             # pass the errors back
             log['stderr'] = stderr
+            import pdb; pdb.set_trace()
```
rhc54: Just curious - what does this do? Probably want the import at the top with the others, and then a comment here that explains what this command is intending to do.

rhc54:
Agreed - we need to get people moving over. I don't know if this impacts anything, but hopefully @jjhursey can comment on it when he has a chance.

jjhursey:
I'll take a look this week - sorry for the delay.

ribab:
@jjhursey Hi Josh, can you test the perl client with my database changes? I have never used the perl client before. I just need testing to see if the perl client still works, not to see if it fills in the new tables.

jjhursey:
I'm not going to have time to test the Perl client for a while - probably a month or more. Now that we have a nice container based installation process we should think about setting up a CI test for the project. But that is for another day.

In any case, we won't be refreshing the current Open MPI community MTT database right away - so this would only impact new MTT server installations. If we need to update the Open MPI community MTT database we will need to do so manually which would be a different effort. Is that something you all need or are you running your own MTT server on-site?

I'm curious what would happen if a person submits results for these new tables, but the server database does not have the tables. Backwards compatibility as we roll forward.

PR #527, opened 2017-05-03
Colon ":" is now Underscore "_" in path

ribab:
This is the fix that addresses @jsquyres concerns in https://github.com/open-mpi/mtt/pull/518

Copytree plugin will write new paths with underscores instead of colons

Stage names have colons in them, and when writing directories with colons, makefiles in those directories have an issue deciphering their path. This commit fixes the problem by replacing ":" with "_" in new directory.

### DIFF
--- a/pylib/Utilities/Copytree.py
+++ b/pylib/Utilities/Copytree.py
@@ -52,7 +52,7 @@ class Copytree(BaseMTTUtility):
             log['stderr'] = "Src directory not specified"
             return
         # define the dst directory
-        dst = os.path.join(testDef.options['scratchdir'], log['section'])
+        dst = os.path.join(testDef.options['scratchdir'], log['section'].replace(":","_"))
         # record the location
         log['location'] = dst
         # perform the copy


pylib/Utilities/Copytree.py
```diff
@@ -52,7 +52,7 @@ def execute(self, log, keyvals, testDef):
             log['stderr'] = "Src directory not specified"
             return
         # define the dst directory
-        dst = os.path.join(testDef.options['scratchdir'], log['section'])
+        dst = os.path.join(testDef.options['scratchdir'], log['section'].replace(":","_"))
```
jsquyres: Oh sorry -- one question: does `.replace(a,b)` replace *all* instances of `a` with `b`?  Or only the first instance?  (sorry, I don't know Python enough to know which way this particular method uses)

ribab: >>> "abcd:efghi:jklmn".replace(":","_")
'abcd_efghi_jklmn'

ribab: It replaces all

jsquyres: 👍

ribab:
I searched the codebase for every place where directories are made.

TestDef.py creates the "mttscratch" directory, and Copytree.py creates the TestGet directory for the test. No other files make directories.

rhc54:
I can only think of one other place where a directory name is created. If you look at the Git plugin, we create a directory name where the cloned repo will go underneath the mttscratch location. I suspect you'll need to do that replace function there too.

ribab:
@rhc54 , I took a look at the Git plugin. It uses the Git repository name instead of using the Stage name for the directory.

rhc54:
Good - thanks for checking!

PR #528, opened 2017-05-03
Fixes #526: Combinatorial now doesn't try to write ENV into temp INI files

### LINKED ISSUE #526: Environment input into INI broke combinatorial executor
Filing an issue before fixing

Issue was that when combinatorial executor writes new INI files, it tries to grab the hidden ENV section. My fix will make the combinatorial INI writer skip the ENV section

ribab:
By skipping ENV section in combinatorial executor, this section (which is created dynamically) is not written to the temporary INI files that is used by the combinatorial executor. Before this fix, an error occurred where combinatorial executor tried to interpolate ${} syntax from environment variables while writing to INI files.


noahv:
Looks good.  Thanks.

PR #531, opened 2017-05-16
RPM to install tests/bat/ folder

ribab:
BAT tests need to be installed with the RPM. This merge will enable that.


noahv:
Looks good thanks.

Issue #532, opened 2017-05-16
default_check_profile.ini BAT test checks for hardcoded kernel version


ribab:
While testing out the Dockerized Jenkins console, the default_check_profile.ini test was failing due to checking for the wrong kernel version, which was hardcoded into the INI file.

Fix is to parameterize kernel version within the INI and within the console and require users to change this parameter to match their machine.

PR #534, opened 2017-05-18
BAT and Jenkins Console parameterized to work on any system

ribab:
* DefaultCheckProfile BAT is parameterized to work on any system
* negative.sh is set to be executable, negative.ini now passes on fail
* Jenkins console configued to set parameters and execute in configurable location


ribab:
This fixes #532 by the way.

PR #538, opened 2017-06-25
Git shallow clone support

amaslenn:
Git [shallow clone](https://git-scm.com/docs/git-clone#git-clone---depthltdepthgt) may help significantly reduce time of clone operation for slow connections. In many case (such as automated testing) full repo history is not required, so this is optimization without side effects.


miked-mellanox:
+1

PR #552, opened 2017-07-29
Topic/stop on fail stderr

noahv:


ribab:
It seems to me that Pull Request #552 contains the changes from #551 and #550 also.

Other than that, trusting that you tested, the changes look good.

My summary of your changes:
* You are appending to library path environment variables
* You have Git plugin create a directory to clone things into
* sequential.py now also prints out stderr when a test fails

ribab:
Consensus is that we merge this pull request and close #551 and #552.

PR #555, opened 2017-08-08
GNU_Install.pm: add "VERBOSE=1" to "make check"

jsquyres:
Per https://www.mail-archive.com/devel@lists.open-mpi.org/msg20334.html.

### DIFF
--- a/lib/MTT/Common/GNU_Install.pm
+++ b/lib/MTT/Common/GNU_Install.pm
@@ -119,7 +119,7 @@ sub Install {
         $ENV{LD_LIBRARY_PATH} = "$config->{libdir}";
     }
 
-    $x = MTT::Common::Do_step::do_step($config, "make check", 1);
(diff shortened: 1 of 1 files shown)


jsquyres:
@noahv I was poking around in `pyclient` and `pylib`, and I don't see any equivalent functionality in the Python MTT client to run `make check`.  Am I missing it somewhere?  If so, I can add the equivalent update here in this PR (to make the Python client run `make check VERBOSE=1`).

jsquyres:
@noahv I'm merging this so that it can be in tonight's MTT runs.  But let me know if I need to make an equivalent PR for the Python client.

noahv:
@jsquyres The python client autotools plugin does not currently support running make check although it should.  I will add an enhancement request for that and include the need for VERBOSE=1.  Thanks for the heads up!

PR #560, opened 2017-08-29
Added allocate/deallocate to WWulf plugin

ribab:
Also fixed variable name conflicts in other files that were caused when adding allocate/deallocate to those plugins


noahv:
The underscore was added to ensure we didn't overwrite log['stdout'] = stdout from the plugins

PR #563, opened 2017-09-04
OMPI: update code for getting version (ep. 2)

amaslenn:


amaslenn:
@jsquyres @alinask @miked-mellanox please review

lib/MTT/Values/Functions/MPI/OMPI.pm
```diff
@@ -167,19 +167,18 @@ sub get_version {
 
     my $match = qr/ompi:version:full:(.*)$/;
 
-    if (not open INFO, "$bindir/ompi_info --parsable|") {
-        open INFO, "$bindir/ompi_info -V|";
+    my @out = `$bindir/ompi_info --parsable`;
+    if ($? >> 8) {
+        @out = `$bindir/ompi_info -V`;
```
miked-mellanox: missing check if ompi_info succeeded

amaslenn: Don't need it. Next loop will match nothing and return it. Very backward compatible :)

miked-mellanox: but no next loop here? it is inside "if"
what do I miss

amaslenn: `for` for `@out` is outside this `if`, closing `}` is on line 174.

PR #566, opened 2017-09-14
Clean handling env vars now prints whole line and ignores commented out env vars

ribab:
This fixes bug where a commented out env variable will not be ignored, now it is ignored like it should be.

Also, more verbose information, such as filename, line number, and whole line is printed out when env variable is missing.


noahv:
Looks good.  Thanks Ricky.

PR #567, opened 2017-09-15
For SLURM plugin, test_list now supports arguments for each test

ribab:
example:
```
test_list = sometest_executable_1.exe list of some arguments
            sometest_executable_2.exe list of some other arguments
            sometest_executable_3.exe list of some more arguments              
```

Also, test_list,fail_tests,fail_returncodes are now deliminated by '\n' rather than ','

I tested this and it works.


noahv:
Looks good.

PR #568, opened 2017-09-15
SLURM plugin 'np' switch correct usage implemented

ribab:
Now 'np' will add '-n' instead of 'N' if srun is used.

Also, 'np' in results log is logged as number of processes, rather than having ambiguity


noahv:
You are on fire!  Nice work.

PR #570, opened 2017-09-19
MTT now exits with error message when no INI is inputted

ribab:
This feature had already been implemented but didn't seem to be working. Basically, add a check to check one more level down in the array and we're good.

args.ini_files is a double array that looks like this:
`[["file1.ini", "file2.ini", "file3.ini", ...]]`

So check on args.ini_files needs to check args.ini_files[0]

### DIFF
--- a/pyclient/pymtt.py
+++ b/pyclient/pymtt.py
@@ -233,7 +233,7 @@ testDef.printInfo()
 
 # if they didn't specify any files, then there is nothing
 # for us to do
-if not args.ini_files:
+if not args.ini_files or not args.ini_files[0]:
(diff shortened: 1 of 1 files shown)


pyclient/pymtt.py
```diff
@@ -233,7 +233,8 @@
 
 # if they didn't specify any files, then there is nothing
 # for us to do
-if not args.ini_files:
+import pdb; pdb.set_trace()
```
noahv: Yeah, this is a typo; must have been left in the code from debug...

noahv:
Please remove the pdb line

ribab:
Line is removed. I am merging thanks guys!

PR #574, opened 2017-09-20
SLURM plugin mpiexec support for the job_name parameter

noahv:

### DIFF
--- a/pylib/Tools/Launcher/SLURM.py
+++ b/pylib/Tools/Launcher/SLURM.py
@@ -322,9 +322,26 @@ class SLURM(LauncherMTTTool):
         skipStatus = int(cmds['skipped'])
         # assemble the command
         cmdargs = [cmds['command']]
-        if cmds['job_name'] is not None:
+
+        # Add support for using job_name with mpiexec
+        if (cmds['command'] == 'mpiexec' or cmds['command'] == 'mpiexec.hydra' or cmds['command'] == 'mpirun') and cmds['job_name'] is not None:
+            if cmds['options'] is None or (cmds['options'] is not None and '-bootstrap slurm' not in cmds['options']):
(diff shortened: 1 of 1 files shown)


rhc54:
I'm a tad concerned here that this option looks very MPICH-specific. Wouldn't it be better to create an MPICH plugin and move such options there?

ribab:
@rhc54 This is tested using Intel MPI, so it is not related to MPICH

rhc54:
The two are largely the same - regardless, my point remains. I don't think we should make the slurm plugin be specific to a particular MPI implementation. We already have MPI plugins - why not put this option over there?

rhc54:
Okay, I think I've studied this enough to see where the confusion lies. You are passing the "job_name" option to SLURM, not "mpiexec", which is fine - has nothing to do with which MPI is being used. However, what I find confusing is why you are mixing the use of mpirun to launch a job into the SLURM launcher plugin?

I was expecting that the SLURM plugin would solely be used to start jobs with "srun" - e.g., you might use the "allocate" entry to get an allocation, and then the "IMPI" plugin to launch the job via "mpiexec". I think you'd be better off to maintain that separation.

noahv:
Hi Ralph, I agree with your comments and will file an enhancement request to create new Intel MPI and MPICH plugins.  

Since this does not modify the default srun usage model we will merge this as so we can continue to make forward progress until those plugins are developed.

rhc54:
Kewl - if it helps, I started writing "allocate" and "deallocate" stages so you could separately get an allocation, run some number of tests, and then deallocate it. The way the code is now, you get an allocation for each test you launch, which seems a little klunky.

If you think that is worthwhile, I will continue working on it and push a PR probably sometime this week.

PR #579, opened 2017-10-02
WWulf and Shell plugins now deallocate on failure

ribab:
Between the allocate and deallocate, these two plugins had the chance to exit
early without deallocating. Now, they will execute a quick "deallocate"
before exiting.


ribab:
@rhc54 I will merge this since it fixes a bug, however yes we should talk about this.

My thought was we only need an allocation for things that need access to the compute nodes. TestGet and TestBuild usually don't need an allocation. Therefore, I thought each plugin should handle their own allocation since not all plugins need an allocation.

PR #584, opened 2017-10-11
Simpler time tracking, and JunitXML time report

ribab:
Rather than adding up all time executions in ExecuteCmd plugin,
this implements a better approach where sequential.py tracks
time execution outside of the plugin, and appends it to the log.

JunitXML plugin is also modified to grab the time info from the log
and report it.


noahv:
Looks good Ricky.  I like how simple this approach is.  Thanks.

PR #585, opened 2017-10-11
Fix syntax error in IPMITool.py

ribab:
ipmiQueue = Queue.Queue(2 * len(controllers))
    creates a syntax error

ipmiQueue = Queue(2 * len(controllers))
    does not create a syntax error

I changed the code to not create the syntax error


noahv:
Looks ok to me.

PR #588, opened 2017-10-11
Shell plugin now doesn't need a parent

ribab:
If no parent is specified, Shell plugin will now use
the scratch directory as the default location to
execute content in.

This allows Shell plugin to be executed in any stage

### DIFF
--- a/pylib/Tools/Build/Shell.py
+++ b/pylib/Tools/Build/Shell.py
@@ -31,6 +31,7 @@ from BuildMTTTool import *
 # @param fail_returncode           Specifies the expected failure returncode of this test
 # @param allocate_cmd              Command to use for allocating nodes from the resource manager
 # @param deallocate_cmd            Command to use for deallocating nodes from the resource manager
+# @param asis_target                    Specifies name of asis_target being built. This is used with \"ASIS\" keyword to determine whether to do anything.
 # @}
 class Shell(BuildMTTTool):
     def __init__(self):
@@ -50,7 +51,7 @@ class Shell(BuildMTTTool):
         self.options['fail_returncode'] = (None, "Specifies the expected failure returncode of this test")
         self.options['allocate_cmd'] = (None, "Command to use for allocating nodes from the resource manager")
         self.options['deallocate_cmd'] = (None, "Command to use for deallocating nodes from the resource manager")
-        self.options['binary'] = (None, "Specifies name of binary being built. This is used with \"ASIS\" keyword to determine whether to do anything.")
+        self.options['asis_target'] = (None, "Specifies name of asis_target being built. This is used with \"ASIS\" keyword to determine whether to do anything.")
         return
 
     def activate(self):
@@ -121,6 +122,7 @@ class Shell(BuildMTTTool):
             return
 
         # get the location of the software we are to build
+        parentlog = None
         try:
             if cmds['parent'] is not None:
                 # we have to retrieve the log entry from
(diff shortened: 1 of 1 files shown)


ribab:
I modified this again to support ASIS better and create directory if used in TestGet. @rhc54 @noahv can you look at it again?

rhc54:
I'm not entirely sure I understand the purpose of as-is-target. Can you perhaps give me an example where this is needed, and why you now have to specify the target when using the ASIS label?

ribab:
Hi @rhc54 the ASIS target is a flexible way to tell whether the shell plugin has already been run.

If in a MiddlewareGet or TestGet stage, then the Shell plugin will create a directory based on the section name, so your ASIS target would be the directory that is created. If the directory already exists, then the Shell plugin would have already been run and ASIS would skip running the plugin again.

If in a MiddlewareBuild or TestBuild stage, then the Shell plugin will build a binary file. The ASIS target would be the name of the binary file, and if it exists then that means the Shell plugin has already been run and ASIS would skip running the plugin again.

rhc54:
I thought that behavior is what ASIS already did, yes? In other words, it looked at what was supposed to be created, and skipped the step if that already exists. I'm trying to understand the need for _requiring_ that the user also provide an explicit target option vs just letting ASIS figure out the default.

Trying to be clear: I would rather not have to supply both a section name, and also a "target" option that is simply a copy of that section name. It seems unnecessary.

ribab:
Makes sense @rhc54 , I will make it so that if you are in a "TestGet" or "MiddlewareGet" section, the Shell plugin will not use the "asis_target" option and instead will check the section name.

If in a build stage though, the "asis_target" is needed.

ribab:
@rhc54 , I made the change just now so that asis_target is only used in MiddlewareBuild and TestBuild stages, and so that ASIS checks if the directory exists in MiddlewareGet and TestGet stages.

Thanks

pylib/Tools/Build/Shell.py
```diff
@@ -132,34 +134,57 @@ def execute(self, log, keyvals, testDef):
...
 
+        # Check to see if this needs to be ran if ASIS is specified
         try:
-            if cmds['asis'] and os.path.exists(os.path.join(location,cmds['binary'])) \
-                            and os.path.isfile(os.path.join(location,cmds['binary'])):
-                testDef.logger.verbose_print("As-Is binary " + os.path.join(location,cmds['binary']) + " exists and is a file")
-                log['location'] = location
-                log['status'] = 0
-                return
+            if cmds['asis'] and (log['section'].startswith("MiddlewareBuild:") or log['section'].startswith("TestBuild:")):
```
ribab: ASIS will look for asis_target if in MiddlewareBuild or TestBuild stage

pylib/Tools/Build/Shell.py
```diff
@@ -132,34 +134,57 @@ def execute(self, log, keyvals, testDef):
...
+                if cmds['asis_target'] is None:
+                    log['status'] = 1
+                    log['stderr'] = "ASIS is used but asis_target option is not specified"
+                    return
+                if os.path.exists(os.path.join(parentloc,cmds['asis_target'])):
+                    testDef.logger.verbose_print("asis_target " + os.path.join(parentloc,cmds['asis_target']) + " exists")
+                    log['location'] = location
+                    log['status'] = 0
+                    return
+            if cmds['asis'] and (log['section'].startswith("MiddlewareGet:") or log['section'].startswith("TestGet:")):
```
ribab: ASIS will look for directory based on section name if in MiddlewareGet or TestGet stage

rhc54:
I think you may have misunderstood me, so let me try to clarify. What I'm recommending is that we always check for the 'target' option. If it isn't present, however, default to using the stage name as the target instead of generating an error.

My logic is based on the original architecture. As you may recall, the plan was that the output from any stage would go into a directory based on the stage itself, plus the name of that stage. One consequence of that architecture is that the user _cannot_ have two stage instances of the same name for the same stage - e.g., you cannot have two Build stages named "foo".

The resulting scratch directory tree looked like this:

                                   <scratch-head>
                                          |
                -------------------------------------------------------
                |                                                     |
              Build                                                  Run
                |                                                     |
         --------------                                         --------------
         |            |                                         |             |
        Foo          Bar                                       Foo           Ding

As you can see, it is allowed that Build and Run stages each have an instance named "Foo", but Build (for example) could not have two instances named "Foo" as their scratch directories would collide.

This is why I questioned the need for the 'target' option - there should never be any question of where the stage output is going. However, if you have use-cases where some ambiguity might exist, then I'm fine with providing the option - so long as we always default to the stage instance name.

Does that help clarify?

pylib/Tools/Build/Shell.py
```diff
@@ -132,34 +134,58 @@ def execute(self, log, keyvals, testDef):
...
+                    else:
+                        testDef.logger.verbose_print("asis_target " + os.path.join(parentloc,cmds['asis_target']) + " does not exist. Continuing...")
+                else: # no asis target, default to check for directory
+                    if os.path.exists(location):
+                        testDef.logger.verbose_print("directory " + location + " exists. Skipping...")
+                        log['location'] = location
+                        log['status'] = 0
+                        return
+                    else:
+                        testDef.logger.verbose_print("directory " + location + " does not exist. Continuing...")
```
ribab: Above is the changes I made to match @rhc54 's description

ribab:
@rhc54 -- that helps clarify. I made changes to match your description.

pylib/Tools/Build/Shell.py
```diff
@@ -132,34 +134,58 @@ def execute(self, log, keyvals, testDef):
...
+            try:
+                parentloc = parentlog['location']
+                location = parentloc
+            except KeyError:
+                log['status'] = 1
+                log['stderr'] = "Location of package to build was not specified in parent stage"
+                return
+        else:
+            try:
+                if log['section'].startswith("TestGet:") or log['section'].startswith("MiddlewareGet:"):
```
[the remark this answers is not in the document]
ribab: If Shell plugin isn't used in the TestGet or MiddlewareGet stage, then the parent location is inherited

pylib/Tools/Build/Shell.py
```diff
@@ -132,34 +134,58 @@ def execute(self, log, keyvals, testDef):
...
-                testDef.logger.verbose_print("As-Is binary " + os.path.join(location,cmds['binary']) + " exists and is a file")
-                log['location'] = location
-                log['status'] = 0
-                return
+            if cmds['asis']:
+                if cmds['asis_target'] is not None:
+                    if os.path.exists(os.path.join(parentloc,cmds['asis_target'])):
+                        testDef.logger.verbose_print("asis_target " + os.path.join(parentloc,cmds['asis_target']) + " exists. Skipping...")
+                        log['location'] = location
+                        log['status'] = 0
```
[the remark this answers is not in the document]
ribab: location is updated already to be section name by that point if in the TestGet or MiddlewareGet stage. Otherwise, the location is inherited by the parent

rhc54:
Ah, I see - thanks for clarifying! All good to go, then

PR #591, opened 2017-10-13
server: Misc changes necessary to port to AWS from IU servers

### LINKED ISSUE #495: Port back AWS changes
We made a number of updates to the MTT server code when Open MPI moved from the IU servers to the AWS servers. Some are portability items, others are logging management.

jjhursey:
* cherrypy: Disable global trial flag
 * Switch postgres serialization option from depricated value
 * reporter: PHP 7 replace split
 * CherryPy misc
   - Add a Get for the /submit interface for debugging
   - Some documentation in the main MTTServer.py
 * Rename config script to template so we don't accidentally commit it
 * Add date string to debug output
 * Add downtime pages, for quick reference
 * Add a helpful error message when gunzip fails
 * Move to using a config file for the cron perl scripts (Add tmpl)

Resolves #495


jjhursey:
I'm not really sure who is able to review this. It's the PHP server code, and I think I'm the only one left that has ever used it.

I'm putting a WIP flag on this for the moment. I'd like to re-install this on the MTT server once it's back up (down currently for a database problem that should resolve on Monday). Just to make sure I didn't miss anything.

server/php/cherrypy/src/webapp/dispatchers.py
```diff
@@ -255,6 +255,20 @@ def _validate_test_run(self, submit_id, metadata, data):
...
+    @cherrypy.tools.json_out()
+    def GET(self, **kwargs):
+        prefix = '[GET /submit/]'
+        self.logger.debug(prefix)
+
+        rtn = {}
+        rtn['status'] = 0
+        rtn['status_message'] = 'Success (submit)'
+
+        return rtn
```
ribab: What's the difference between GET() on line 261 and GET() on line 275?

jjhursey: Ah yes. I already brought that over. I'll push a fix today.

jjhursey: fixed in latest push. Thanks

jjhursey:
I'm installing the new version now. I found a handful of items that need to be pushed back to this PR. Will do that today.

jjhursey:
The cron perl scripts have some repeated code. I didn't have the cycles to put that in a library.

This is the version running on the AWS system right now. I think this is ready to go.

jjhursey:
Any objections to merging this soon? It's what we are currently running, and I'd like to get the AWS current to the HEAD of master so there are no changes out there.

PR #592, opened 2017-10-18
Adding plugin to build hostfile based on nodelist

ribab:
Plugin is "pylib/Tools/Build/Hostfile.py"

Plugin takes in two inputs:
 - hostfile
 - nodelist

nodelist can be in format:
    node[01-03],node05,node[07-08,10-12],node14,node15

hostfile location is a relative location


noahv:
Nice work Ricky!

PR #595, opened 2017-10-20
Removed unused options from all plugins

ribab:
Several options were unimplimented in plugins, so I removed them.

We can add them again as we create functionality.

These removed options include:
 - force
 - logfile
 - submit_group_results
 - save_stdout_on_success
 - save_stdout_on_pass
 - report_after_n_results
 - timeout


rhc54:
Just FYI: you'll eventually have to restore most of those options. What you will discover is that you can't just collect all the results and then hit the database with them at the end - if you hit an error somewhere, then you lose all the intermediate ones. It also can overwhelm the database when the commit is too large.

Likewise, it often helps with debugging to redirect output to a logfile, and to save stdout so you can see what the test actually did.

I recognize that we (probably me) didn't fully implement all the options due to lack of time. Still, just be aware that we are going to need them.

Issue #597, opened 2017-11-02
New v3 tables cause Perl client to fail


jjhursey:
The recent commit (https://github.com/open-mpi/mtt/commit/348b56b7e4e01d0d1189416500c0d31aa84b5c41) introduced new tables to the v3 database schema. We received a report that this causes the Perl client to fail to submit to the database because of the default value set. Email snippet below:

> the perl client cannot always correctly talk to the server with the schema-v3. For example, I have a lot of problems with the bios not being specified during the submission, the server therefore uses the default value (-38) which creates an error.

noahv:
How can I help with this?  Do we need to create a new pull request for https://github.com/gvallee/mtt/tree/gv_dbfix?

gvallee:
I let you guys decide, i could not sign-off the PR, no idea why it is not working. It would be great to have the patch included but there is not much i can personally do at this point.

jsquyres:
What PR are you referring to?  There are no open PRs.

Are you asking for some help with git to get this straightened out?

gvallee:
I close the PR since nothing was happening and I have limited amount of time to spend on this. I do not want to have a PR open for weeks, it is just not productive. It was PR #598

noahv:
We can discuss this today in the MPI Testing Tool Developer Workgroup.   Please let me know if you would like to join us; 11 AM PST.

jsquyres:
Looking at #598, it looks like run-of-the-mill-Git-newbie kinds of issues.  Probably fairly easy to sith down with @gvallee and fix (e.g., on a webex or somesuch).

gvallee:
I already spent time reading the documentation, trying to fix the problem. and it did not help. I don’t have the time to sit in a meeting for a 3-line modification. It is not strait forward, I let someone else get the patch in. Sorry.

jsquyres:
Intel / @ribab it looks like your commit is the one that broke things. Can you please fix things?  It’s been nearly a week since this bug was reported.   Steal from @gvallee’s branch if you need to - he figured out the problem but ran into got issues. If his fix is correct, please make sure to give him credit (i.e., make him the commit author).

rhc54:
It has already been fixed, guys: https://github.com/open-mpi/mtt/commit/57555e71c1273341d452e59cc6c6425f6b8a6815

Committed yesterday afternoon.

jsquyres:
Excellent -- thanks!

jjhursey:
This was fixed by PR #600

PR #598, opened 2017-11-02
Fix issue #597. Modification to schemas-v3 so the Perl client can sub…

gvallee:
…mit results.

I ran a basic test and it seems to be okay and fix the problem with the Perl client.

This also raises a good question: which fields are mandatory? I do not believe that data such as bios should be, i can foresee some facilities refuse to share that type of information.


jjhursey:
@gvallee can you add a signoff to this commit (`git commit --amend -s` should do it)?

Fixes https://github.com/open-mpi/mtt/issues/597

gvallee:
Done but it does not seem to fix the problem.

gvallee:
No idea why it is not working. I did something wrong the first time but should have worked the second time.

PR #601, opened 2017-11-08
Build::Noop: support redefinition of install dir

amaslenn:
Signed-off-by: Andrey Maslennikov <andreyma@mellanox.com>

### DIFF
--- a/lib/MTT/Test/Build.pm
+++ b/lib/MTT/Test/Build.pm
@@ -454,6 +454,10 @@ sub _do_build {
         $ret->{start_timestamp} = timegm(gmtime());
         $ret->{refcount} = 0;
 
+        if ($ret->{noop_installdir}) {
+            $ret->{srcdir} = $ret->{noop_installdir};
+        }
+
         if (!defined($ret->{test_result})) {
             $ret->{test_result} = MTT::Values::FAIL;
         }
--- a/lib/MTT/Test/Build/Noop.pm
+++ b/lib/MTT/Test/Build/Noop.pm
@@ -12,13 +12,25 @@ package MTT::Test::Build::Noop;
 
 use strict;
 
+use MTT::Values;
+
 #--------------------------------------------------------------------------
 
 sub Build {
+    my ($ini, $mpi_install, $config) = @_;
+
+    # In case of tests located in different folder and no much space available
+    # on target scratch, it can be useful to point another location of prebuilt
+    # tests. To make it more safe to use name contains package prefix `noop_`
+    # to work with this module only.
+    my $user_defined_install_dir = Value($ini, $config->{full_section_name}, "noop_installdir");
+
     my $ret;
     $ret->{test_result} = MTT::Values::PASS;
     $ret->{exit_status} = 0;
     $ret->{result_message} = "Success";
+    $ret->{noop_installdir} = $user_defined_install_dir if $user_defined_install_dir;
+
     return $ret;
 } 
 


amaslenn:
@jsquyres @miked-mellanox @alinask please review.

lib/MTT/Test/Build.pm
```diff
@@ -454,6 +454,10 @@ sub _do_build {
         $ret->{start_timestamp} = timegm(gmtime());
         $ret->{refcount} = 0;
 
+        if ($ret->{noop_installdir}) {
+            $ret->{srcdir} = $ret->{noop_installdir};
```
miked-mellanox: what is a use-case?
why srcdir=installdir?

amaslenn: This is required to pass through a path to binaries to Run section (`srcdir` is used to search for binaries for a test). I used `installdir` to highlight that it is a directory which should contain binaries, not sources.

jsquyres:
Can you please explain what this PR is for / how it will be used?

amaslenn:
Sure. The goal is to use MTT to build tests separately and then use these prebuilt tests in actual runs. This will decrease run duration (test-get/test-build are skipped) and increase host usage efficiency.

MTT has the only `install path`, so I pass a path to prebuilt tests location in affected sections only.

rhc54:
FWIW: you could just start using the Python client which already has all this functionality, Just a suggestion.

amaslenn:
@rhc54 thanks! Could you please point exact parameter? I'm not familiar with python client.

rhc54:
I believe Intel has pushed all the Python client docs up to the wiki, but I couldn't immediately find them.

@noahv Has that happened yet? Any way we can get the docs up there?

@amaslenn The Python client replaces the Perl one and has a lot of advanced features, including flags to indicate that you should reuse the results of a section if they already exist.

amaslenn:
@rhc54 I need to take a look, thanks again.
Anyway, we won't migrate to python client right now, so this PR is still relevant.

miked-mellanox:
+1

jsquyres:
Wait -- I still don't understand the use case for this PR.

Also, it seems like you're totally violating the abstraction boundary by putting plugin-specific logic in the core `Build.pm`.

I was literally about the "request changes" on this PR when @miked-mellanox merged it.  😦   Please revert.

miked-mellanox:
@jsquyres - sorry, pressed merge too fast. The use-case is here:  https://github.com/open-mpi/mtt/pull/601#issuecomment-342812952

jsquyres:
I see that comment, but I don't fully understand it.  Just give me a little time to grok...

jsquyres:
Here's my cognitive dissonance -- perhaps you can correct my thinking:

- There's `Noop` modules for MPI get/install and Test get/build.
- The `Noop` modules were intended for just that -- no operations.  They were intended just to setup internal data structures so that later MTT stages could act as if the prior stages had already occurred.
- This change added a param to `Build/Noop` to allow the tests to be in a different location.  That seems like a fine idea, but it violated some abstractions and overloaded field names to put in a hackaround (i.e., overrode srcdir with an installdir -- that would be very confusing to track down if there's ever a bug in this code path).  That seems dangerous, and could potentially cause someone quite a bit of time to track that down someday.

Is that incorrect?

It would seem like a better solution would be:

* Migrate to the Python client.  As Ralph mentioned, we're trying to get rid of the Perl client (stop adding new features to it!).  Sure, it's easy to "just add this one feature -- I swear I'll switch to the Python client someday," but we need to stop doing that (because that's what we'll *always* do).
* Or create a solution in the Perl client that does not break abstractions.  If the test build data structure assumes that the test sources and executables are in the same directory and you need them to be in different directories, then introduce a new field that differentiates between source dir and executable dir, and have `Build/Noop.pm` properly reset the executable dir field to the user-specified location (and make sure that it works with the other `Build` modules).

Honestly, if we're going to go to the Python client, then we should *go* to the Python client.

rhc54:
It's still a little early out here, but hopefully @noahv or @ribab will be in shortly and can address the documentation question.

Continuing FWIW: in the Python client, you simply add ```ASIS``` to the name of a section to tell MTT to execute the section if the end product isn't present, and skip the section if it is already there. The client automatically tracks end product locations, and so any subsequent steps will get the right input.

noahv:
We have not yet added documentation for the python client ASIS.

We recently updated the support for ASIS in PR #558

rhc54:
Have you pushed the rest of the Python client docs? I couldn't find them when scanning the wiki

noahv:
No we have not done that yet

PR #604, opened 2017-11-09
Build::Noop: support external test location

amaslenn:
Original discussion: https://github.com/open-mpi/mtt/pull/601.

### DIFF
--- a/lib/MTT/Test/Build.pm
+++ b/lib/MTT/Test/Build.pm
@@ -557,6 +557,10 @@ sub _do_build {
         $report->{submit_id} = $submit_id;
         $ret->{submitl_id} = $submit_id;
 
+        if ($ret->{installdir}) {
+            $report->{installdir} = $ret->{installdir};
+        }
+
         # Submit it!
         my $serials = MTT::Reporter::Submit("Test Build", $simple_section, $report);
 
--- a/lib/MTT/Test/Build/Noop.pm
+++ b/lib/MTT/Test/Build/Noop.pm
@@ -12,13 +12,24 @@ package MTT::Test::Build::Noop;
 
 use strict;
 
+use MTT::Values;
+
 #--------------------------------------------------------------------------
 
 sub Build {
+    my ($ini, $mpi_install, $config) = @_;
+
+    # In case of tests located in different folder and no much space available
+    # on target scratch, it can be useful to point another location of prebuilt
+    # tests.
+    my $install_dir = Value($ini, $config->{full_section_name}, "installdir");
+
     my $ret;
     $ret->{test_result} = MTT::Values::PASS;
     $ret->{exit_status} = 0;
     $ret->{result_message} = "Success";
+    $ret->{installdir} = $install_dir if $install_dir;
+
     return $ret;
 } 
 
--- a/lib/MTT/Test/Run.pm
+++ b/lib/MTT/Test/Run.pm
@@ -429,7 +429,10 @@ sub _do_run {
         $ini->val($mpi_details_section, "network$suffix");
 
     # Go to the right dir
-    MTT::DoCommand::Chdir($test_build->{srcdir});
+    # 'installdir' is defined by build module, so it has higher priority
+    my $test_bin_dir = $test_build->{installdir} ? $test_build->{installdir}
+                                                 : $test_build->{srcdir};
+    MTT::DoCommand::Chdir($test_bin_dir);
 
     # Set the PATH and LD_LIBRARY_PATH
     if ($mpi_install->{bindir}) {


amaslenn:
@jsquyres please comment.
I used better naming to better reflect my intent.

miked-mellanox:
+1

amaslenn:
\+ @alinask

lib/MTT/Test/Build.pm
```diff
@@ -416,6 +416,12 @@ sub _do_build {
     } elsif ($config->{srcdir}) {
         $ret = MTT::Module::Run("MTT::Test::Build::$config->{module}",
                                 "Build", $ini, $mpi_install, $config);
+        if ($ret->{noop_test_bin_dir}) {
+            # At Build step 'srcdir' is used to point to location of binaries.
+            # For Noop module it is possible to redefine it with external test
+            # location.
+            $config->{srcdir} = $ret->{noop_test_bin_dir};
```
jsquyres: Thanks for this new PR.

But this still has two of the same problems I mentioned in the last PR:

1. You're using a module-specific value here in the core `Build.pm`.  This is an abstraction violation --
 `Build.pm` should be completely agnostic of all Build modules.
1. You're overriding `srcdir` with a value that is *not* the source directory.  This is very confusing -- i.e., a value named `$config->{srcdir}` *isn't* the source directory.  If you want to allow the test executables to be in a different directory than the source dir, then make a new `$config->{bindir}` field (that presumably defaults to the same value of `$config->{srcdir}`).  Use `$config->{bindir}` as appropriate here in `Build.pm` (and/or elsewhere -- perhaps `Run.pm` will need to use it too...?  You'll need to check).  Then `Build/Noop.pm` can just override `$config->{bindir}`.

amaslenn:
@jsquyres don't get me wrong, but I didn't invent `build`/`src` misleading:
https://github.com/open-mpi/mtt/blob/71b3fef96f285751833e844e83266f907985aefa/lib/MTT/Test/RunEngine.pm#L224-L226

Agree with abstraction violation. Will try to find a better solution.

jsquyres:
Heh -- I think you're trying to use the past against me.  😄 

FWIW: I think that cause is more about initially setting the build dir to be the same as whatever the source dir was set to -- not trying to overload 2 different things into a single value.

amaslenn:
@jsquyres please review. Please also note, that I like this new solution :)

lib/MTT/Test/RunEngine.pm
```diff
@@ -222,7 +222,7 @@ sub RunEngine {
         $run->{test_build_simple_section_name} = $test_build->{simple_section_name};
 
         # Setup some globals
-        $MTT::Test::Run::test_build_dir = $test_build->{srcdir};
+        $MTT::Test::Run::test_build_dir = $test_build->{bindir};
```
jsquyres: This might be a similar case as before: the *build* directory is different than the *source* or the *install* directory.  This variable is returned by the funclet `test_build_dir()`, but note that we already have a `test_executable_dir()`.

The build directory is *usually* the source directory, but it could also be a VPATH build tree.

But I don't think it's correct to default it to the bin directory.

lib/MTT/Test/Build/Noop.pm
```diff
@@ -12,13 +12,24 @@ package MTT::Test::Build::Noop;
...
+
 #--------------------------------------------------------------------------
 
 sub Build {
+    my ($ini, $mpi_install, $config) = @_;
+
+    # In case of tests located in different folder and no much space available
+    # on target scratch, it can be useful to point another location of prebuilt
+    # tests.
+    my $test_bin_dir = Value($ini, $config->{full_section_name}, "test_bin_dir");
```
jsquyres: I'm sorry to be picky, but the more I think about this, the more I think we should call it the "install" dir, because there's typically these generic directories:

* Source directory
* Build directory
* Install directory

The word "install" is a little better because the installed files may or may not be binaries (they'll almost certainly be executables of some flavor, but not necessarily binaries).

lib/MTT/Test/Build/Noop.pm
```diff
@@ -12,13 +12,24 @@ package MTT::Test::Build::Noop;
...
+
     my $ret;
     $ret->{test_result} = MTT::Values::PASS;
     $ret->{exit_status} = 0;
     $ret->{result_message} = "Success";
+    $ret->{bindir} = $test_bin_dir if $test_bin_dir;
```
jsquyres: We want `$ret->{bindir}` to exist but set to `undef` if it wasn't set, right?

I.e., the infrastructure should always be able to assume that the `bindir` (or `installdir`, per the previous comment) field *exists*, even if it's effectively empty.

lib/MTT/Test/Build/Noop.pm
amaslenn: Considering only these three options, it should be `install`. Should it be `external_installdir` or just `installdir` is OK?

lib/MTT/Test/Build/Noop.pm
amaslenn: Not necessarily. I added a check into Run.pm:
https://github.com/open-mpi/mtt/pull/604/files#diff-bd02d29cb58bf48db2504639705bcd7aR433, which set `binder = srcdir` if first one missed. Otherwise I need to add `$ret->{bindir} = under` into every build module for consistency.
Anyway, see next comment.

lib/MTT/Test/RunEngine.pm
amaslenn: OK, need to think. The previous chunk+comment are related to this one, most likely will be changed together.

lib/MTT/Test/Build/Noop.pm
jsquyres: My $0.02: I think just `installdir` is fine -- no need for `external` in there.

lib/MTT/Test/RunEngine.pm
jsquyres: Cool.  Thanks!

jsquyres:
Looks great!  Thanks for your patience.

Can you squash down that last commit?  Having just the changes in a single commit made it great to review, but for the permanent record / commit log, let's have it so that the right code was just in the first commits and there was no need for a commit to fix the prior commits.  Make sense?

amaslenn:
Clean history is important, will squash in a minute.

PR #606, opened 2017-11-29
Added Copytree plugin switch for preserve_symlinks.

noahv:
Default value is to not preserve symlinks and copy the contents instead.


ribab:
Looks good to me

Issue #610, opened 2018-01-08
Cherrypy RESTful API queries have errors with some fields


ribab:
Fix is in place and ready to commit.

Issue was that a step was missing in translating between field display names and DB names

PR #611, opened 2018-01-08
Add trans. between field disp-name and DB name

ribab:
Translation added between field display-naems and DB-names
on the CherryPy RESTful Query API server code
This fixes bug #610


ribab:
I had a mistake when creating this pull request. Creating another one

PR #622, opened 2018-02-12
"deactivate" ran for plugins when interrupted

ribab:
When code execution is interrupted by ctrl+c or by some other means, the plugins "deactivate" function is called, which can give plugins a chance to clean up.

The cleanup is implemented for the Harasser plugin

### DIFF
--- a/pylib/Tools/Build/Shell.py
+++ b/pylib/Tools/Build/Shell.py
@@ -343,9 +343,9 @@ class Shell(BuildMTTTool):
         cfgargs = shlex.split(cmds['command'])
 
         if 'TestRun' in log['section'].split(":")[0]:
-            harass_exec_ids = testDef.harasser.start(log, testDef)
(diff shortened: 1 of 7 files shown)


ribab:
I fixed the typo in the ALPS plugin

ribab:
@rhc54 can you take a look?

ribab:
@rhc54 
**"you simply are detecting that harassers were still running when the plugin stopped, and if so, you clean them up"**
Yes. You got it right.

**"The rest of the changes were just moving the harasser log into the main log instead of the log for each stage."**
No. I removed the log from some harasser functions because it wasn't being used.

The change is a general interrupt handler that could be used for cleaning up any plugin that needs cleaning up with an unexpected interrupt. (In this case, Harasser is cleaning up lingering harassers)

The change also ensures that the reporter stage will be executed when an unexpected interrupt occurs.

rhc54:
Okay, I now understand - looks fine. Thx!

PR #625, opened 2018-02-21
Generate and email report for no test results

amaslenn:


lib/MTT/Reporter/HTMLReport.pm
```diff
@@ -238,11 +242,24 @@ sub _summary_report
 
     ";
 
+    my $ignore_min_succ = 0;
+    my $extra_info = "";
+    if (not $total_tests) {
+        $extra_info = "<h2>Extra info</h2><span style='color:red'>No tests were executed.</span>";
+        if (MTT::Util::time_to_terminate()) {
+            $extra_info .= "\n<span>Terminate file found.</span>";
```
jsquyres: Do you want to put this extra info in (that the terminate file was found) regardless of whether any tests were run or not?

amaslenn: No, I use it as a tip to help understand why no there were no tests.

jsquyres: What I mean is: the Terminate message is relevant regardless of whether any tests were run or not.  E.g., I periodically `touch terminate-file` to drain my SLURM queues of MTT jobs in an orderly fashion -- the currently running jobs will stop in the middle (and therefore have `$total_tests>0`.  All jobs after that will have `$total_tests==0` (because they'll notice the terminate file immediately and not actually run any tests).  The point is that both cases can (and do) happen -- so it is probably worthwhile to pull the "Terminate file found" message out of the check for how many `$total_tests` have run.

For debugging purposes, you'll still get your Extra Info from `No tests were run`, but it's *also* useful to know that the overall MTT job completed because of a termination file.

Make sense?

amaslenn: Got you. Make sense, will update.

amaslenn:
@jsquyres is anything else required here?

jsquyres:
No, as far as I'm concerned, you can merge when ready.

amaslenn:
I don't have merge permissions :)

@miked-mellanox could you help, please?

jsquyres:
@amaslenn Oh!  Sorry about that -- didn't realize.  Glad @miked-mellanox took care of it.

PR #631, opened 2018-03-01
pyclient:make trial variable more like perl client

### LINKED ISSUE #630: Python client treats trial variable oddly
The IU reporter in the python client handles the presence of the ```trial``` variable oddly.  If one just includes ```trial``` in the ini file, the python client thinks this means its trial run.  You can set trial to anything, e.g. 0, and you'll still get a trial run.  This seems to violate the rule of least astonishment, esp. for MTT perl client savy users.

hppritcha:
fixes #630

### DIFF
--- a/pylib/Stages/Reporter/IUDatabase.py
+++ b/pylib/Stages/Reporter/IUDatabase.py
@@ -141,7 +141,7 @@ class IUDatabase(ReporterMTTStage):
         metadata['mtt_client_version'] = '4.0a1'
         metadata['platform_name'] = self._extract_param(testDef.logger, 'MTTDefaults', 'platform')
         trl = self._extract_param(testDef.logger, 'MTTDefaults', 'trial')
-        if trl:
+        if trl and trl.lower() != "false":
             metadata['trial'] = 1
         else:
             metadata['trial'] = 0


pylib/Stages/Reporter/IUDatabase.py
```diff
@@ -142,7 +142,7 @@ def execute(self, log, keyvals, testDef):
         metadata['platform_name'] = self._extract_param(testDef.logger, 'MTTDefaults', 'platform')
         trl = self._extract_param(testDef.logger, 'MTTDefaults', 'trial')
         if trl:
-            metadata['trial'] = 1
+            metadata['trial'] = trl
```
jjhursey: Should we try to normalize this into a `0` or `1` before to gets to the server?
`metadata['trial'] = 1 if trl else 0`
(check my Python to make sure that ternary expression will work)

ribab:
The reason why setting "trial" to anything will cause it to act as if "trial" were set to True is this:

Setting a variable in MTT will set it in string format.   "False" != False in python

I suggest to add this to the "if" check:

```
        trl = self._extract_param(testDef.logger, 'MTTDefaults', 'trial')
        if trl and trl.lower() != "false":
            metadata['trial'] = 1
        else:
            metadata['trial'] = 0
```

hppritcha:
I like @ribab 's suggestion and will modify per his suggestion

hppritcha:
@ribab check now.  I went with your suggestion.
@jjhursey

PR #632, opened 2018-03-05
Working watchdog plugin executable by INI

ribab:
Watchdog plugin made to be executable from an INI file.
Semaphores added to prevent watchdog from throwing interrupt while program can't catch the interrupt.

### DIFF
--- a/pylib/Tools/Executor/sequential.py
+++ b/pylib/Tools/Executor/sequential.py
@@ -66,16 +66,24 @@ class SequentialEx(ExecutorMTTTool):
         status = 0
         only_reporter = False
         stageOrder = testDef.loader.stageOrder
+
+        # Holding a semaphore while in transition between plugins
+        # so async threads don't interrupt in the wrong context
+        testDef.plugin_trans_sem.acquire()
+
         for step in testDef.loader.stageOrder:
             for title in testDef.config.sections():
                 if only_reporter and step != "Reporter":
                     continue
                 try:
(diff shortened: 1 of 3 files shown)


rhc54:
I think it would really help if you put some explanation at the top of the sequential plugin explaining what all these calls to "acquire" mean. As things stand, I have no idea what is going on (since I am not intimately familiar with the Semaphore library), and therefore have no way to review this PR.

ribab:
@rhc54 The semaphores are there to prevent the watchdog timer from generating an exception while the main thread is not within the context that catches exceptions.

I'll add the comments to describe their function in the code

pylib/Tools/Executor/sequential.py
```diff
@@ -66,16 +66,24 @@ def execute(self, testDef):
         status = 0
         only_reporter = False
         stageOrder = testDef.loader.stageOrder
+
+        # Holding a semaphore while in transition between plugins
+        # so async threads don't interrupt in the wrong context
+        testDef.plugin_trans_sem.acquire()
```
ribab: Hi @rhc54 -- I added this comment to describe the semaphores more. Does this help?

ribab:
Hi @rhc54 -- I added this comment to describe the semaphores more. Does this help?

```
+        # Holding a semaphore while in transition between plugins
+        # so async threads don't interrupt in the wrong context
+        testDef.plugin_trans_sem.acquire()
```

PR #633, opened 2018-03-12
Put a link to terminate file in Extra info section

amaslenn:


amaslenn:
@jsquyres please review

amaslenn:
@jsquyres could you please take a look?

lib/MTT/Reporter/HTMLReport.pm
```diff
@@ -250,8 +250,10 @@ sub _summary_report
     if (not $total_tests) {
         push @extra_info_parts, "<span style='color:red'>No tests were executed.</span>";
     }
-    if (MTT::Util::time_to_terminate()) {
-        push @extra_info_parts, "<span>Terminate file found.</span>";
+
+    my $term_file = MTT::Util::find_terminate_file();
+    if (defined $term_file) {
+        push @extra_info_parts, "<span>Terminate <a href='$term_file'>file</a> found.</span>";
```
jsquyres: Why the hyperlink?

I'm not sure it's a good assumption that the terminate file will be on the same filesystem as the web server.  If you want the report to contain the contents of the termination file, it is probably much safer to include it by value in `@extra_info_parta`.

amaslenn: The content of a terminate file can be huge. Putting it directly to report may introduce a lot of mess, that is why I preferred a hyperlink. The worst case is this case - invalid link. Is it a user scenario to create a termination file somewhere outside?

jsquyres: Actually, this HTML report is emailed out -- so there's almost zero possibility that the link won't be broken (because the recipient will be viewing their mail on their local mail client, which almost certainly has no connection to the server where the termination file resides).

Perhaps you can add some params to the HTMLReport reporter that specifies a URI prefix for where the termination file can be found...?  I.e., if the user specified that, then you can put in a link.  Otherwise, don't put in a link.

jsquyres:
@amaslenn When you squash to 1 commit, if you don't have merge permission and need me to merge it, post another comment here because github does not send notifications for git force pushes on PRs.

Thanks!

amaslenn:
Squashed! @jsquyres please merge.

PR #634, opened 2018-03-13
Add --duration and --loopforever switches to MTT

ribab:
With --duration switch, MTT will have a maximum duration for tests to run.
This duration applies to all tests being run by MTT, even if multiple tests
are ran at a time. (If duration ends test will interrupt and return success)

With --loopforever switch MTT will continue to run forever until interrupted.

The two switches work together to have a test keep running over and over for a specified amount of time.

Allocation and deallocation commands were fixed so that deallocation happens if
the execution was interrupted by the duration timing out (or by another interruption)


ribab:
By the way, I tested this patch to make sure the duration switch caused an interrupt, and also to make sure that "deallocate" is called when the interrupt happens.

ribab:
@rhc54 that is true.

PR #637, opened 2018-03-20
Comment out the md5/sha1 check for now

jjhursey:
Disable the md5 and sha1 checks since Amazon S3 cannot support these files
in a way that doesn't expose a timing hole which can cause the tarball to
download with an old md5. So for now disable the checks, which are likely
not needed.


bwbarrett:
https://github.com/open-mpi/mtt/pull/638 is probably a better solution...

jjhursey:
https://github.com/open-mpi/mtt/pull/638 is better

PR #641, opened 2018-03-22
OMPI snapshot: add way to specify MPI name/version

hppritcha:
The Open MPI project has a particular way they like
to specify the MPI name and version.  Go to

https://mtt.open-mpi.org/

and click on the "summary" key.

This commit adds the ability to specify the "MPI name"
and "MPI version" within the OMPI snapshot plugin.

The autotools plugin is modified to check if
the mpi_name and version key's are already present
in the global log dictionary of dictionaries.  If
present, don't set in the autotools plugin.


ribab:
@hppritcha -- can we merge? Or did you want to check on something first?

PR #647, opened 2018-04-22
HTML Report: small enhancements 

amaslenn:


amaslenn:
@jsquyres please review. CC @alinask.

Issue #648, opened 2018-04-24
Python client executecmd doesn't always return full output 


ribab:
if process finishes before python can grab all output the executecmd plugin fails to grab all output. 

I fixed this in a patch i will submit

ribab:
This is fixed by #649

PR #649, opened 2018-04-24
ExecuteCmd plugin now shows all output

ribab:
Before ExecuteCmd would stop displaying output when the program it is executing finishes. Now it displays all output.


ribab:
This fixes #648

noahv:
Looks good.  Thanks.

PR #650, opened 2018-04-24
Topic/hostfile fraction of nodes

ribab:
Hostfile to dynamically generate list of nodes

This list can be a fraction of all nodes, or a fraction of a subset

**Merging this will also merge the fix for ExecuteCmd #649 (if not already merged)

### DIFF
--- a/pylib/Utilities/ExecuteCmd.py
+++ b/pylib/Utilities/ExecuteCmd.py
@@ -110,24 +110,32 @@ class ExecuteCmd(BaseMTTUtility):
                 reads = [p.stdout.fileno(), p.stderr.fileno()]
                 ret = select.select(reads, [], [])
 
+                stdout_done = True
+                stderr_done = True
+
                 for fd in ret[0]:
                     # if the data
                     if fd == p.stdout.fileno():
-                        read = p.stdout.readline().rstrip()
+                        read = p.stdout.readline()
                         if read:
+                            read = read.rstrip()
                             testDef.logger.verbose_print('stdout: ' + read.decode("utf-8", errors='replace'))
                             if merge:
                                 stderr.append(read.decode("utf-8", errors='replace'))
                             else:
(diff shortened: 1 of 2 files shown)


pylib/Utilities/ExecuteCmd.py
```diff
@@ -110,24 +110,32 @@ def execute(self, options, cmdargs, testDef):
...
-                        read = p.stderr.readline().rstrip()
+                        read = p.stderr.readline()
                         if read:
+                            read = read.rstrip()
                             testDef.logger.verbose_print('stderr: ' + read.decode("utf-8", errors='replace'))
                             stderr.append(read.decode("utf-8", errors='replace'))
+                            stderr_done = False
 
-                if p.poll() != None:
+                if stdout_done and stderr_done:
```
rhc54: I'm probably just missing it, but I'm not sure I understand this completion test. If I read either stdout or stderr, then you set the corresponding flag to False - which means I will exit this loop. But that means I won't have read the other fd should something have come out that way. Is that the intended behavior - i.e., you only wait for the first one to get output and then quit?

rhc54: Oh wait - it's the other way around. You only exit if neither one has something to read. I gather that p.stdout.readline and p.stderr.readline are blocking calls then?

rhc54: Guess what I'm saying is that I'm having a real problem understanding this logic. I gather that the p.poll test was a problem somehow??

ribab: Yes, the p.poll test was a problem.
I tried using a debugger to see what p.poll returns even before all output was read using readline(), and it returns 0 in the debugger

ribab: I only quit if both stdout and stderr have no output (not even a \n newline as output)

rhc54: ok - thanks!

PR #654, opened 2018-05-30
JunitXML more readable xml output

ribab:
stdout and stderr had unreadable raw python list format
New format is newline delimited which is more readable


ribab:
I tested this and it works 👍

rhc54:
Trivial change - feel free to merge.

PR #658, opened 2018-06-09
Fixed Harasser plugin for None in join_timeout.

noahv:
Added backtrace support to exception handler.


ribab:
We tested this small change and it works -- merging

PR #662, opened 2018-06-18
Revert "lib: added JSON lib"

rhc54:
Reverts open-mpi/mtt#646

This code is licensed as GPL - we are reverting as soon as we realized it.


jsquyres:
See https://github.com/open-mpi/mtt/pull/646#issuecomment-398209268 for details.

Issue #435, opened 2016-05-31
python module loader could be a little smarter


hppritcha:
Currently the 

``` python
LoadClasses.load
```

method isn't so smart.  It gets confused if one executes the pymtt.py client from anywhere but
$MTT_HOME/mtt/pyclient directory.  

It would be nice if the load method could be a little smarter and leverage the MTT_HOME environment variable.

PR #664, opened 2018-06-23
Refactor plugin loading to use pathlib

PeterGottesman:
This fixes an issue where relative plugin directories would not load
correctly. Furthermore, it simplifies the flow of the program
significantly, making it much more readable.
The changes to TestDef.py also fix an issue where the error reported was
not always the error which occurred. In addition to catching
non-existent directories, it caught any error which occurred in
LoadClasses.load() and gave unhelpful responses.
Per a conversation with @rhc54 , plugins in the given plugin
directories will be loaded, as opposed to just their subdirectories.

pathlib and hostlist were added to pyenv.txt. pathlib was included as
it is not standard in python <= 3.4. hostlist was an existing
requirement which was not in pyenv.txt.

I chose to use pathlib as it is much more concise and readable for some operations (such as getting all files in a tree matching a basic pattern) than os.path. The documentation for the pre-python3.4 version is available at https://pathlib.readthedocs.io/en/pep428/, and the post-3.4 documentation is on https://docs.python.org.

Tested with python 3.6.5 and python 2.7.9

Signed-off-by: Peter Gottesman <pgottesm@cisco.com>

### DIFF
--- a/pylib/System/TestDef.py
+++ b/pylib/System/TestDef.py
@@ -24,6 +24,7 @@ from yapsy.PluginManager import PluginManager
 import datetime
 from distutils.spawn import find_executable
 from threading import Semaphore
+from pathlib import Path
 
 is_py2 = sys.version[0] == '2'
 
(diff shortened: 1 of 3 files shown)


rhc54:
@ribab @noahv would you please review these changes?

pylib/System/TestDef.py
```diff
@@ -257,19 +258,17 @@ def loadPlugins(self, basedir, topdir):
...
-                filez = os.listdir(dirPath)
-                for file in filez:
-                    file = os.path.join(dirPath, file)
-                    if os.path.isdir(file):
-                        self.loader.load(file)
-            except:
-                if not self.options['ignoreloadpatherrs']:
-                    print("Plugin directory",dirPath,"not found")
-                    sys.exit(1)
+                if not Path(dirPath).exists():
```
jsquyres: Minor quibble: will `Path(dirPath).exists()` ever throw an exception?  If not, does it belong inside this `try` block?  More specifically: is the `self.loader().load(...)` the only thing that this `try` block is really protecting?

hppritcha:
Peter, could you add fixes #435 to your commit message?

hppritcha:
@PeterGottesman

hppritcha:
I'd like to double check this with my python setup, but can't get to it till Wednesday.

PeterGottesman:
@hppritcha Amended commit message, squashed changes. Let's wait to merge until you have tested on Wednesday.

hppritcha:
Do I still need to set MTT_HOME environment variable?

PeterGottesman:
The MTT_HOME environment variable needs to be set when you are not in the project root and didn't call pymtt with an absolute path. I did not change the logic around that, in short I just fixed the issue where it would fail when called from the root directory.
I don't know where you are calling from, but it will likely not hurt to set MTT_HOME.

ribab:
I tested this code with a simple test and it worked. Plugins loaded just fine.

PR #666, opened 2018-07-06
setup travis to run single-node bat and setup website on gh-pages branch

Jagaskak:
Run bat located in the tests/bat directory (these only require one node to execute). Also setup travis to push changes to GitHub pages website with every successful merge as well as to create the website itself. Also took out unnecessary lines from negative.ini.


.travis.yml
```diff
@@ -34,27 +34,51 @@ dist: trusty
...
       - doxygen-gui
       - graphviz
+
+      # install MPI libs
+      - openmpi-bin
+      - mpich
+      - libmpich-dev
+      - libopenmpi-dev
+
+      # install SLURM
```
rhc54: Why install slurm?

.travis.yml
```diff
@@ -34,27 +34,51 @@ dist: trusty
...
       - doxygen
       - doxygen-doc
       - doxygen-latex
       - doxygen-gui
       - graphviz
+
+      # install MPI libs
+      - openmpi-bin
+      - mpich
+      - libmpich-dev
```
rhc54: Why install mpich?

docs/README.md
```diff
@@ -0,0 +1,224 @@
+What is this software?
+----------------------
+
+This is the MPI Testing Tool (MTT) software package.  It is a
```
rhc54: I believe we moved away from calling this the "MPI" testing tool when we started building the Python client as it became more of a generic capability. I've heard suggestions for "Middleware" testing tool, though I think we should throw it open to broader consideration. Certainly "MPI" is no longer really appropriate.

.travis.yml
```diff
@@ -34,27 +34,51 @@ dist: trusty
...
+  - $TRAVIS_BUILD_DIR/pyclient/pymtt.py --verbose --base-dir=$TRAVIS_BUILD_DIR/pylib $TRAVIS_BUILD_DIR/tests/bat/default_check_profile.ini
+
+before_deploy:
   - cd docs && doxygen Doxyfile
 deploy:
   provider: pages
   skip_cleanup: true
   github_token: $GITHUB_TOKEN
-  local_dir: $TRAVIS_BUILD_DIR/docs/html
+  local_dir: docs/
```
rhc54: I don't know Travis at all - is there some place I should see where it creates pages in the gh-pages branch per the title on this PR?

.travis.yml
Jagaskak: I installed SLURM to test if I could run the bat that tested the multi-node capabilities without spinning up a cluster. I forgot to remove all the libraries I installed when testing out the other bat. I'll remove these since they are not necessary for the single node bat to execute.

.travis.yml
hppritcha: I'm also curious why this change at line 54.

rhc54:
@Jagaskak This looks okay to me - I'll have the community address the name question. Please answer @hppritcha question before committing, but otherwise should be fine.

.travis.yml
Jagaskak: The local-dir flag uploads the contents of that directory to the gh-pages branch. But it assumes a path relative to the current directory, so giving the absolute path made Travis complain because it kept wanting to change dir to an invalid path (i.e. /home/travis/build/Jagaskak/mtt/home/travis/build/Jagaskak/mtt/docs/html)

Travis also happens to overwrite the contents of the gh-pages branch when you upload a file, which is why I added the files to build the website to the docs folder and uploaded the entire file. 

On line 55, which you can't see here, I set the keep-history flag to true, so that Travis doesn't overwrite the history of the branch as well.

Issue #680, opened 2018-07-26
Python client log interpolation of byte strings


PeterGottesman:
In TestDef.py's `fill_log_interpolation` method, the byte string type imported from python 3 is not handled. Here is the traceback with some extra added debug information about the failure:

```
Setting basestr: TestRun_IBMInstalledOMPI.testresults.0.stderr to b'[Errno 2] No such file or directory'
Sublog type is: <class 'future.types.newbytes.newbytes'>
LOGGING results for Reporter: text file backup
=======================================
Exception was raised: <type 'exceptions.TypeError'> option values must be strings
=======================================
Traceback (most recent call last):

  File "/home/pgottesm/PyMTT/mtt/pylib/Tools/Executor/sequential.py", line 127, in execute
    testDef.configTest()

  File "/home/pgottesm/PyMTT/mtt/pylib/System/TestDef.py", line 553, in configTest
    self.fill_log_interpolation(e['section'].replace(":","_"), e)

  File "/home/pgottesm/PyMTT/mtt/pylib/System/TestDef.py", line 527, in fill_log_interpolation
    self.fill_log_interpolation("%s.%s" % (basestr, k), v)

  File "/home/pgottesm/PyMTT/mtt/pylib/System/TestDef.py", line 533, in fill_log_interpolation
    self.fill_log_interpolation("%s.%d" % (basestr, i), v)

  File "/home/pgottesm/PyMTT/mtt/pylib/System/TestDef.py", line 527, in fill_log_interpolation
    self.fill_log_interpolation("%s.%s" % (basestr, k), v)

  File "/home/pgottesm/PyMTT/mtt/pylib/System/TestDef.py", line 524, in fill_log_interpolation
    self.fill_log_interpolation(basestr, sublog.encode("utf-8", "ignore"))

  File "/home/pgottesm/PyMTT/mtt/pylib/System/TestDef.py", line 522, in fill_log_interpolation
    self.config.set("LOG", basestr, sublog.replace("$","$$"))

  File "/home/pgottesm/PyMTT/venv/lib/python2.7/site-packages/backports/configparser/__init__.py", line 1238, in set
    _, option, value = self._validate_value_types(option=option, value=value)

  File "/home/pgottesm/PyMTT/venv/lib/python2.7/site-packages/backports/configparser/__init__.py", line 1221, in _validate_value_types
    raise TypeError("option values must be strings")

TypeError: option values must be strings
```

[...]

PR #682, opened 2018-07-26
Removed unicode check from fill_log_interplation function

ribab:
The check will be handled implicitly in the "else",
and the check for unicode conflicts with using python3

This is related to #680


ribab:
All the other times "encode" is used is right before writing to a file.

I checked, and python3 can write it's binary data-type that results from the encoding to a file.

The purpose of the encoding in the first place was that some test content had output in a weird encoding, and there was an error displaying the output when using Jenkins.

--Ricky

PR #683, opened 2018-07-30
Ensure spastic.o is cleaned up after the compiler test

PeterGottesman:


pylib/Utilities/Compilers.py
```diff
@@ -184,6 +184,7 @@ def check_compile(self, testDef, macro, c_code, compiler):
 
         # cleanup the test
         os.remove("spastic.c")
+        os.remove("spastic.o")
```
jsquyres: Do you need to test to ensure that these files exist?  (I know nothing about the rest of this `.py` file -- just asking if it's *always* true that both `spastic.c` and `spastic.o` will exist)

PeterGottesman: Yep, forgot to commit before I pushed. Should be fixed now

PR #684, opened 2018-07-31
Minor updates

jsquyres:
A series of small commits for minor fixes that Cisco came across while working to bring up the Python client on our internal MTT cluster.

We tested these with Python 3.6.5 and 2.7.9.

@ribab Two notes for you:

1. We fixed a minor bug where child processes could be orphaned.
1. I see that the logging *always* encodes the output in utf-8.  Is there a reason for that?
    * I asked @rhc54 that question earlier today and he thought there was a reason, but he couldn't remember what it was.
    * @PeterGottesman and I talked about this on the phone; he initially thought it was a Python 2-vs-3 compatibility issue, but we determined that it isn't.
    * I *assume* that the utf-8 encoding isn't (or: is no longer) necessary, so one of the commits in here takes it out.  We noticed this because it makes weird output with python 3 -- a bunch (but not all) of the MTT output is enclosed in `b''`.  Let me know if we need to do something different.

### DIFF
--- a/pylib/Utilities/ExecuteCmd.py
+++ b/pylib/Utilities/ExecuteCmd.py
@@ -90,6 +90,7 @@ class ExecuteCmd(BaseMTTUtility):
 
         # it is possible that the command doesn't exist or
         # isn't in our path, so protect us
+        p = None
         try:
             if time_exec:
                 starttime = datetime.datetime.now()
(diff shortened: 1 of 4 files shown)


jsquyres:
@ribab Any thoughts on this PR, perchance?

pylib/Utilities/ExecuteCmd.py
```diff
@@ -118,19 +119,19 @@ def execute(self, options, cmdargs, testDef):
...
-                                stdout.append(read.decode("utf-8", errors='replace'))
+                                stdout.append(read)
                             stdout_done = False
                     elif fd == p.stderr.fileno():
                         read = p.stderr.readline()
                         if read:
-                            read = read.rstrip()
-                            testDef.logger.verbose_print('stderr: ' + read.decode("utf-8", errors='replace'))
-                            stderr.append(read.decode("utf-8", errors='replace'))
+                            read = read.rstrip().decode('utf-8', errors='replace')
```
ribab: Hi @jsquyres ,
I checked the PR and I it looks good to me. The code looks more clean.
I saw you didn't actually take out the utf-8 decoding though, and instead just moved it to another location.

jsquyres: In this case, yes, because reading from stdout/stderr results in a byte stream, not a string.  `decode()` here will convert those bytes into a Python string.

PR #687, opened 2018-08-09
Fixed return values of ExecuteCmd, and with how TextFile reporter interprets them

### LINKED ISSUE #685: Utilities/ExecuteCmd.py inconsistent return values
In a code review, we noticed that `ExecuteCmd::execute()` (https://github.com/open-mpi/mtt/blob/master/pylib/Utilities/ExecuteCmd.py#L39) sometimes returns the following for `stdout` and `stderr`:

* `None`
* A string
* A list of strings

It looks like different callers expect different things throughout the code base.  There does not appear to be uniform testing for `None`, for example (which generates errors if you try to execute with `--dry-run`, for example).

Perhaps this function could either:

1. Always return either `None` or a list (even if the list is of length 1 if there's only a single string)
1. Always return a list (even if the list is of length 0, 1, or >1)

I do not fully understand the code base, so I did not want to try to update all the places that call `ExecuteCmd::execute()`.

ribab:
Fixes #685

Now if an error happens in the code, the return value will still be a list

### DIFF
--- a/pylib/Stages/Reporter/TextFile.py
+++ b/pylib/Stages/Reporter/TextFile.py
@@ -55,6 +55,12 @@ class TextFile(ReporterMTTStage):
             print(prefix + line)
         return
 
+    def _print_stderr_block(self, name, lines, tabs=1):
+        if lines:
+            print("\t"*tabs,"ERROR ({name})".format(name=name), file=self.fh)
+            for l in lines:
+                print("\t"*(tabs),"   ",l, file=self.fh)
+
     def execute(self, log, keyvals, testDef):
         testDef.logger.verbose_print("TextFile Reporter")
         # pickup the options
@@ -87,13 +93,10 @@ class TextFile(ReporterMTTStage):
                 except KeyError:
                     pass
                 if 0 != lg['status']:
-                    try:
-                        print("\tERROR:",lg['stderr'], file=self.fh)
-                    except KeyError:
-                        try:
-                            print("\tERROR:",lg['stdout'], file=self.fh)
-                        except KeyError:
-                            pass
+                    if "stderr" in lg:
+                        self._print_stderr_block("stderr", lg['stderr'], tabs=1)
+                    if "stdout" in lg:
+                        self._print_stderr_block("stdout", lg['stdout'], tabs=1)
             except KeyError:
                 pass
             try:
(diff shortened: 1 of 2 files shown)


pylib/Stages/Reporter/TextFile.py
```diff
@@ -88,10 +88,16 @@ def execute(self, log, keyvals, testDef):
                     pass
                 if 0 != lg['status']:
                     try:
-                        print("\tERROR:",lg['stderr'], file=self.fh)
+                        print("\tERROR:", file=self.fh)
+                        if lg['stderr']:
```
jsquyres: I don't think you need these `if` statements because the result will *always* be a list, right?  (although the list might be empty)

I think you can just go directly to the `for l in ...` statement.

jsquyres: Actually, I think you need to restructure this a little.  You don't need the try blocks any more, and you want to check to see if you want to print the ERROR heading first.  Perhaps something like this:

```python
def _print_error_block(name, lines):
    if len(lines) > 0:
        print("\tERROR ({name})".format(name=name), file=self.fh)
        for l in lines:
            print(l, file=self.fh)

if 0 != lg['status']:
    _print_error_block(lg['stderr'])
    _print_error_block(lg['stdout'])
```

PeterGottesman: I would also just add a `if 'stderr' in lg:` and similar for stdout. Otherwise, agreed with these points.
Also, mind the missing `name` arg 👍

jsquyres: Hah ya; that's what one gets for typing code out in a web browser and not testing it.  😉

ribab:
I made a few changes based on your review @jsquyres , thanks

pylib/Stages/Reporter/TextFile.py
```diff
@@ -87,13 +93,10 @@ def execute(self, log, keyvals, testDef):
                 except KeyError:
                     pass
                 if 0 != lg['status']:
```
PeterGottesman: if you change this to `if 'status' in lg and lg['status'] != 0:`, and then you can remove the enclosing try/except to save on some indentation and improve readability.
Generally, it is probably a good idea to minimize the try/excepts all over the place for the sake of maintainability and readability.

PeterGottesman: If `status` is guaranteed to be in `lg`, you don't even need to do that, you can just leave the if as-is and remove the try/except.

ribab: I understand you don't like the overuse of try/accepts, however I am not adding these to the code they already existed beforehand.

I will try to reformat the code to remove try/accepts along with the other changes I made to ExecuteCmd and TextFile

rhc54: Just to be fair, I'm almost certainly the one who wrote that code. IIRC, the try/accept code was in there because at that time we hadn't required the values to be present, and hadn't settled on them being a list. Things were pretty unstable back then 😄 

I have no objection to removing them now that things have settled down.

PR #690, opened 2018-08-14
Cleanup the TestDef discovery code

### LINKED ISSUE #673: Python client TestDef.py location resolution
The following is the bug I mentioned at last week's telecon, it is also the actual source of #435.

The code I am referring to here is in pyclient/pymtt.py at lines 157-227
## Description of problem:
 - Logic to find the MTT TestDef Class varies based on where code is run from
 - Sets two variables, `basedir` and `topdir`
 - basedir: Directory which should contain the TestDef.py and LoadClasses.py files
 - topdir: Directiory which should contain the default plugins
 - Comment from the code:
```
    # Try to find the MTT TestDef class. Try several methods:

    # 1. If they specified the directory, check it and underneath
    #    it, if necessary
    # 2. See if we are in the MTT home directory by looking for
    #    the MTT plugin directory directly underneath us
    # 3. Check to see if MTT_HOME is set in the environment
    # 4. If $0 is a path, try seeing if that is the place.
```

  - The actual cases which are handled, in order:
    1. If `basedir` is set by --base-dir on launch
       - Check `basedir` for TestDef.py
       - Check `basedir/System`
       - Check `basedir/pylib/System`
       - If found:
         - basedir is the directory containing TestDef.py
         - ***topdir is whatever --base-dir was defined as***
    2. elif "TestDef.py" exists in the current directory
       - topdir = basedir = './' -> 'pylib/System'
       - ***No plugins will be found***
    3. elif "pylib" exists in the current directory
       - basedir = './pylib/System'
       - topdir = './pylib/'
    4. else
       - Try getting the MTT_HOME env variable and using that
         - This is done with a try/except instead of checking if
         if the key exists
         - basedir = '$MTT_HOME/pylib/system'
         - topdir = '$MTT_HOME/pylib'
       - Check if this was called as an absolute path
         - path = dirname $0 (i.e. pyclient/)
         - basedir = 'path/pylib/System' -> 'pyclient/pylib/System'
         - topdir = 'path/pylib' -> 'pyclient/pylib'
         - ***No plugins will be found***
## Possible Solutions:
~~In my eyes, the best solution is to do away with this code entirely.
There are so many assumptions made about the directory structure that
it certainly makes sense to hard code the location of topdir in relation to
the top level directory of the repository.
As for basedir, it is unneeded in my opinion. Since the files in basedir are
mandatory for running the program, they should be in the pyclient directory.
This would allow us to simply `import TestDef` without having to do any class
loading.
While that seems like the logical solution to me,~~ another option is to keep the
existing structure and refactor this code to be consistent ~~and effective~~.
This would be fairly easy to implement, especially if making reasonable
assumptions about the directory structure. Moreso using pathlib's glob
functionality.

@ribab @hppritcha What are your thoughts on this? I don't want to just dive in, let's make sure it is going to work as intended and cover any use case before we change it.

Edit: I think it is also worth discussing the order in which the different cases are handled. One example is that if the environment variable MTT_HOME is set, it is probably not one of the last things we want to check. My proposed order would be:
1. The cli arguments
1. MTT_HOME env variable
1. The canonicalized directory of $0
1. pwd, although I am not sure it is necessary with the prior.

rhc54:
Per telecon of 8/15/2018, simplify the TestDef discovery procedure by
requiring that the user set MTT_HOME in their environment.

Fixes #673 

### DIFF
--- a/pyclient/pymtt.py
+++ b/pyclient/pymtt.py
@@ -153,6 +153,13 @@ except KeyError:
     print("to the top-level directory of your MTT installation")
     sys.exit(1)
 
+# check to see if it is an absolute path, as we require
+if not os.path.isabs(mtthome):
+    print("MTT_HOME environment variable:")
+    print("    ", mtthome)
+    print("is not an absolute path")
+    sys.exit(1)
+
 # check to see if MTT_HOME exists
 if not os.path.exists(mtthome):
     print("MTT_HOME points to a non-existent location:")
--- a/pyclient/pymtt.py
+++ b/pyclient/pymtt.py
@@ -176,11 +176,22 @@ if not os.path.exists(topdir) or not os.path.isdir(topdir):
     sys.exit(1)
 
 # set basedir and check for existence
-basedir = os.path.join(mtthome, "pylib", "System")
+if args.basedir and not os.path.isabs(args.basedir):
+    print("The basedir cmd line option is not an absolute path:")
+    print("   ", args.basedir)
+    print("Please correct")
+    sys.exit(1)
+
+basedir = args.basedir or os.path.join(mtthome, "pylib", "System")
 if not os.path.exists(basedir) or not os.path.isdir(basedir):
-    print("MTT_HOME points to a location that does not\ninclude the \"pylib/System\" subdirectory:")
-    print("   ", basedir)
-    print("does not exist. Please correct")
+    if basedir == args.basedir:
+        print("The basedir cmd line option points to a location that does not exist:")
+        print("   ", basedir)
+        print("Please correct")
+    else:
+        print("MTT_HOME points to a location that does not\ninclude the \"pylib/System\" subdirectory:")
+        print("   ", basedir)
+        print("does not exist. Please correct")
     sys.exit(1)
 
 # if they want debug, set the logging level
--- a/pyclient/pymtt.py
+++ b/pyclient/pymtt.py
@@ -144,87 +144,37 @@ debugGroup.add_argument("--trial",
                       help="Use when testing your MTT client setup; results that are generated and submitted to the database are marked as \"trials\" and are not included in normal reporting.")
 args = parser.parse_args()
 
+# check to see if MTT_HOME has been set - we require it
+try:
+    mtthome = os.environ['MTT_HOME']
+except KeyError:
+    print("MTT_HOME could not be found in your environment")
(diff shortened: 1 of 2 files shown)


rhc54:
@ribab I fear that someone will have to update the Travis tests to reflect these changes

rhc54:
woohoo! fixed Travis

pyclient/pymtt.py
```diff
@@ -144,87 +144,37 @@
...
 args = parser.parse_args()
 
+# check to see if MTT_HOME has been set - we require it
+try:
+    mtthome = os.environ['MTT_HOME']
+except KeyError:
+    print("MTT_HOME could not be found in your environment")
+    print("Python client requires that this be set and point")
+    print("to the top-level directory of your MTT installation")
+    sys.exit(1)
```
PeterGottesman: I think it would make sense to ensure MTT_HOME is an absolute path.
I would go with forcing MTT_HOME to be an absolute path, as opposed to expanding it ourselves.
i.e
```python
if not os.path.isabs(mtthome):
    print("MTT_HOME environment variable is not an absolute path")
    sys.exit(1)
```

PeterGottesman: Otherwise, this looks great. I think this is a good, clean solution to #673.

PeterGottesman:
Do we need to notify users before we merge in this change?

pyclient/pymtt.py
rhc54: good suggestion - I added it

pyclient/pymtt.py
```diff
@@ -144,87 +144,44 @@
...
+# set topdir and check for existence
+topdir = os.path.join(mtthome, "pylib")
+if not os.path.exists(topdir) or not os.path.isdir(topdir):
+    print("MTT_HOME points to a location that does not\ninclude the \"pylib\" subdirectory:")
+    print("   ", topdir)
+    print("does not exist. Please correct")
+    sys.exit(1)
+
+# set basedir and check for existence
+basedir = os.path.join(mtthome, "pylib", "System")
```
PeterGottesman: This is ignoring the fact that basedir might be set by `--base-dir`. Maybe this should be
```python
basedir = args.basedir or os.path.join(mtthome, "pylib", "System")
```

rhc54: okay, added some checks and supported use of base-dir

PR #691, opened 2018-08-16
Fix wrong path for loading testdef and loadclasses

ribab:
The path was changed to be absolute for loading these two classes before,
but the absolute path was assuming these 2 classes are in the pylib folder, but they
are actually in pylib/System folder

This is to fix that

### DIFF
--- a/pyclient/pymtt.py
+++ b/pyclient/pymtt.py
@@ -201,7 +201,7 @@ if (args.debug):
 # load the "testdef" Test Definition class so we can
 # begin building this test
 try:
-    m = imp.load_source("TestDef", os.path.join(basedir, "TestDef.py"))
+    m = imp.load_source("TestDef", os.path.join(basedir, "System/TestDef.py"))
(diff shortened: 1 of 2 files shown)


ribab:
This is related to https://github.com/open-mpi/mtt/pull/690

PeterGottesman:
It seems to be correct already:
https://github.com/open-mpi/mtt/blob/ec79f906ab49855c2e7b7f5c15fe8756b816030d/pyclient/pymtt.py#L185
That seems to be confirmed by the failed CI:
```
FileNotFoundError: [Errno 2] No such file or directory: '/home/travis/build/open-mpi/mtt/pylib/System/System/TestDef.py'
```
What is the full command you ran that caused your error? And what is your MTT_HOME?

ribab:
Ok I agree. The issue with what I was running was the basedir directory that I was setting conflicted with what it expected.

I have always set `--base-dir` to point to `mtt/pylib`

PR #692, opened 2018-08-22
Fixes #674 -- scratch for writing reporter files

### LINKED ISSUE #674: Textfile reporter writing to arbitrary locations
When the python client is executing a section, if the section fails for some reason the results file will be written to the current working directory - i.e. wherever the failure occurred.
Additionally, it seems that when there is no tests to run it is just being written to $HOME.

ribab:
TextFile plugin and JunitXML plugin now write to the pre-defined scratch directory

### DIFF
--- a/pylib/Stages/Reporter/JunitXML.py
+++ b/pylib/Stages/Reporter/JunitXML.py
@@ -57,7 +57,8 @@ class JunitXML(ReporterMTTStage):
         cmds = {}
         testDef.parseOptions(log, self.options, keyvals, cmds)
         if cmds['filename'] is not None:
-            self.fh = open(cmds['filename'], 'w')
+            self.fh = open(cmds['filename'] if os.path.isabs(cmds['filename']) \
+                           else os.path.join(cmds['scratch'],cmds['filename']), 'w')
         if testDef.options['description'] is not None:
             print(testDef.options['description'], file=self.fh)
             print(file=self.fh)
--- a/tests/bat/default_check_profile.sh
+++ b/tests/bat/default_check_profile.sh
@@ -10,5 +10,5 @@ logfile=${LOGFILE:? LOGFILE not set}
 # Check that the logfile has correct version string
 # Remove the line in the logfile that was used for passing in KERNEL_RELEASE environment variable
 # The logfile will live at the top of the <scratchdir>/<testname> directory structure
-cat ../../$logfile | grep -v KERNEL_RELEASE | grep $kernel_release
+cat $logfile | grep -v KERNEL_RELEASE | grep $kernel_release
 exit $?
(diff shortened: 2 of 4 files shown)


ribab:
TravisCI fails because this changes the location that files are being reported to.

I need to change the travisci test as a part of this pull request so that way travisci will pass.

tests/bat/default_check_profile.sh
```diff
@@ -10,5 +10,5 @@ logfile=${LOGFILE:? LOGFILE not set}
 # Check that the logfile has correct version string
 # Remove the line in the logfile that was used for passing in KERNEL_RELEASE environment variable
 # The logfile will live at the top of the <scratchdir>/<testname> directory structure
-cat ../../$logfile | grep -v KERNEL_RELEASE | grep $kernel_release
+cat ../$logfile | grep -v KERNEL_RELEASE | grep $kernel_release
```
jsquyres: Just curious -- if `$logfile` is being set by the user, why isn't it set to the correct logfilename such that you have to prefix it with `..`?

(after thinking about this for a minute) ...oh, is `$logfile` a basename?  If so, it might be easier / more robust to just set `$logfile` to an absolute pathname, and avoid exactly this kind of issue in the future.

Just a thought.

pylib/Stages/Reporter/JunitXML.py
```diff
@@ -57,7 +57,7 @@ def execute(self, log, keyvals, testDef):
         cmds = {}
         testDef.parseOptions(log, self.options, keyvals, cmds)
         if cmds['filename'] is not None:
-            self.fh = open(cmds['filename'], 'w')
+            self.fh = open(os.path.join(cmds['scratch'],cmds['filename']), 'w')
```
PeterGottesman: Would it be worth it to check if `cmds['filename']` is an absolute path before appending it to the scratch directory?

ribab: I am assuming that cmds['scratch'] is already an absolute path when that is set by TestDef

tests/bat/default_check_profile.sh
ribab: That sounds good

pylib/Stages/Reporter/JunitXML.py
PeterGottesman: Right, but if the user sets the reporter's file to `/home/user/not/in/scratch/junit.xml` and scratch is at `/home/user/scratch`, this will output it to `/home/user/scratch/home/user/not/in/scratch/junit.xml`. So I am asking if it is worth it to check  if the filename is already an absolute path.

If it should _just_ be a filename and not a path to a file then it should be fine left as-is IMO. Just wanted your input there.

ribab: I see

I will add a check for abs path before joining the paths then

PR #700, opened 2018-08-24
Fixes #696 Sequential.py now fixes output format of plugins

### LINKED ISSUE #696: mttresults.txt file has stderr in single-column mode
When I hit an error in executing the Python client, the mttresults.txt file contains the following (after output from various stages):
```
	 ERROR (stderr)
	     P
	     r
	     i
	     o
	     r
	      
	     d
	     e
	     p
	     e
	     n
	     d
	     e
	     n
	     t
	      
	     s
	     t
	     e
	     p
	      
	     f
	     a
	     i
	     l
	     e
	     d
	      
	     -
	      
	     c
	     a
	     n
	     n
	     o
	     t
	      
	     p
	     r
	     o
	     c
	     e
	     e
	     d
```
We obviously aren't handling the returned error output quite correctly.

ribab:
type for stdout and stderr is now required to be a list instead of a string

This fix will find when a plugin uses a string, and will convert it to a list

This way we don't have to rewrite every single plugin

### DIFF
--- a/pylib/Tools/Executor/sequential.py
+++ b/pylib/Tools/Executor/sequential.py
@@ -23,6 +23,11 @@ from yapsy.PluginManager import PluginManager
 
 from ExecutorMTTTool import *
 
+try:
+    basestring
+except:
+    basestring = str
+
 # Theory of Operation
 #
 # The sequential executor executes a single, ordered pass thru the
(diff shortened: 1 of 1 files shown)


rhc54:
Fixes #696

PR #701, opened 2018-08-24
Fixes #694 Recreate env_modules_python.py when link is broken

### LINKED ISSUE #694: env_modules_python.py isn't recreated when link is broken
After a system  upgrade, the link MTT creates to `env_modules_python.py` broke for us, and MTT doesn't recreate the file if it already exists. This generates a confusing error that is hard to debug.

Temporary solution (which was already done) for us was to remove mttscratch and regenerate

However real solution is to have MTT recreate that file `env_modules_python.py` if link is broken.

ribab:


rhc54:
Fixes #694

Issue #703, opened 2018-08-25
Do not deploy github pages if there were no changes


jsquyres:
(I asked this in mtt-devel but got no reply, so I figured I'd ask here)

It looks like every time there is a PR merged, there's a new `gh-pages` deployment.  This is generally fine, except that sometimes the deployments are not as a result of any actual documentation changes, but just changes in the timestamps of the documentation.  I.e., no doc changes occurred, but the doc generator ran, it updated the timestamps, and therefore those timestamp changes were recorded as a git diff, which then resulted in another `gh-pages` deployment.

Can such spurious commits / deployments be avoided?

jsquyres:
FWIW, I don't think it's a matter of not invoking Travis -- I think it's fine to invoke Travis every time.  What would be nice is if Travis runs the generator, do the `git commit`, and then perhaps take a cursory examination of the `git diff` and see "oh, this just has timestamp updates" and therefore decide not to `git push`.

I don't know if I have the cycles, either.

rhc54:
I guess it begs this question:

> Git only sees the content of the file to determine if it is changed. It is not concerned with the timestamp if the content is the same. Make sure there really isn't any change in the content, including line endings.

So it sounds like the doc generator runs, but there isn't any change? If all that changes is the timestamp, then there shouldn't have been a diff, yes?

I guess I'm confused as to how just timestamp updates resulted in a diff being committed at all.

rhc54:
Hmmm - I wonder why that information even needs to be committed? Maybe the problem would go away if we removed that file (or at least the timestamps in it) from the repo?

jsquyres:
I think those timestamps are just part of the files that Doxygen generates.

I think it's ok to generate those timestamps -- it shows that the files are current, after all.  It would just be nice if we can somehow determine that the timestamps are the only thing in the diff, and therefore it's not worth pushing.

ribab:
I think the solution to this is:

Only run gh-pages travisci **IF** something in `mtt/docs` changes

PR #704, opened 2018-08-28
Try deploying gh-pages only if change to docs dir

rhc54:

### DIFF
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+set -ev
+testchange=`git diff --exit-code -- docs > /dev/null`
+if [$testchange]; then
+  deploy:
+    provider: pages
+    skip_cleanup: true
+    github-token: $GH_TOKEN
+    local_dir: docs/
+    keep-history: true
+    on:
+      branch: master
+fi
+
--- a/scripts/deploy.sh
+++ b/scripts/deploy.sh
@@ -1,7 +1,11 @@
 #!/bin/bash
-set -ev
+
+# check to see if the docs directory changed - if not,
+# then git will return a 0 exit code. If it did change,
+# then git will return a 1. Only deploy the docs directory
+# if a change occurred.
 testchange=`git diff --exit-code -- docs > /dev/null`
-if [$testchange]; then
+if [ $? -ne 0 ]; then
   deploy:
     provider: pages
     skip_cleanup: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,14 +67,7 @@ script:
 
 before_deploy:
   - cd docs && doxygen Doxyfile
-deploy:
-  provider: pages
-  skip_cleanup: true
-  github-token: $GH_TOKEN
-  local_dir: docs/
-  keep-history: true
-  on:
-    branch: master
+script: ./scripts/deploy.sh
 after_success:
 env:
   matrix:


rhc54:
@ggouaillardet Does this look right to you?

rhc54:
Fixes #703

scripts/deploy.sh
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+set -ev
+testchange=`git diff --exit-code -- docs > /dev/null`
```
jsquyres: Is the rationale that if there's an actual diff in the `docs` directory, then something meaningful changed, and therefore it's worth going through with the deploy?

If so, this might be worth a comment, because Future Jeff / Future Ralph / Future Someone might wonder why this two-step process is here.

rhc54: Per @ribab:

> Only run gh-pages travisci IF something in mtt/docs changes

scripts/deploy.sh
```diff
@@ -0,0 +1,14 @@
...
+if [$testchange]; then
```
jsquyres: I don't think $testchange will be loaded with 0 or 1.

I think you actually want:

```sh
# --quiet inhibits "git diff" output and implies --exit-code
git diff --quiet -- docs
if test $? -eq 1; then
```

..but then again, I'm slightly confused because this does not appear to be a pure Bourne shell script...?

scripts/deploy.sh
rhc54: I can add a comment - my question to @ggouaillardet was: does this work the way I think it should?

rhc54:
FWIW: the logic of that script is extracted directly from the Travis documentation 😄

scripts/deploy.sh
ggouaillardet: Should the test be `if [ $? -ne 0 ]` ? `$testchange` should always be empty.

rhc54: okay, I changed it - but then we have to also remove the "set -ev" line else we will exit if the command returns a non-zero status code, and Travis is adamant that we not call exit.

rhc54: FWIW: my test showed that testchange was also either equal to 0 or 1, not empty

scripts/deploy.sh
```diff
@@ -0,0 +1,14 @@
...
+  deploy:
```
ggouaillardet: All I can think of is this is not bash shell (the name and syntax suggests this should be a bash shell).

I will double check the Travis syntax tomorrow.

Btw, what is the rationale for this PR ? Is this a simple optimization (e.g. it does not fix anything) to save Travis resources ?

jsquyres: @ggouaillardet I asked a question on the MTT devel list to see if we could not re-deploy every time there's a commit to MTT.  It causes a git push + email in the docs tree, and the only thing it contains is a timestamp update in the latex source.  E.g.: https://github.com/open-mpi/mtt/commit/41bbb3c184a1fa4e796bce338862a8ec39b7bd7f

rhc54: Yeah, I'm not convinced this is actually worth all this angst, but was willing to spend five minutes on it just to keep the git repo from unnecessarily growing. If it goes back/forth much more, though, I'll drop it as not worth the time.

jsquyres:
BTW, is the `latex` dir part of the `docs` dir?  I ask because the meaningless timestamp update occurs in the `latex` dir.

rhc54:
Danged if I know. @ggouaillardet Here is the link to the Travis doc:

https://docs.travis-ci.com/user/customizing-the-build/#implementing-complex-build-steps

jsquyres:
Hah!  Yeah, I agree it's not worth a huge amount of time.

But... @ggouaillardet if you could spend 5 minutes on it tomorrow, that would then be a global total of 10 minutes of each of our time, and if that carries this over the finish line, #winning.  😉

rhc54:
I'm open to whatever solution (including doing nothing) someone wants to pursue 😄

ggouaillardet:
@rhc54 Any reason why you closed this PR ?

Right now, doing nothing is the best option. The rationale is the latex file contains some time stamps, so there is always something to push. (It looks like a latex bug/lack of feature, that is likely fixed in the latest version).

I do not think the desired behavior can be achieved with some configuration only.
I would advocate to write a custom script, such as the one in Stack Overflow.
If `git push —force ...` is not acceptable on the `gh-pages` repo, then some extra steps are required.
I found an example on SO, and I can make a proof-of-concept if needed.

rhc54:
I closed it because you indicated it was wrong, so no point in retaining it. If you believe you have a way of solving it, feel free to pursue it. Just don't burn a ton of time on it as it isn't something that critical.

jsquyres:
@ggouaillardet If you're interested...

I think this PR had the right idea: 

* Doxygen generates several different flavors of the same docs. 
* Therefore, if there's changes in one of them, there's changes in *all* of them.
* But there's a bug/feature that the latex flavor *always* changes (it updates timestamps).
* So it might be sufficient to check to see if there's any changes in the HTML docs (which are assumedly in a different directory than the latex docs).  If `git diff --quiet` returns a nonzero exit status in the HTML docs, then there's actually changes in the docs.  If it returns zero, there's no meaningful changes.
* If there *are* meaningful changes, then we should run the deploy.  If not, then don't bother deploying.

ggouaillardet:
@rhc54 @jsquyres I think I now got the rationale right. The missing piece of the puzzle - from my point of view - is we need to (sort of) `git diff docs/html` instead of `git diff docs`, at least as long as the latex file contains some timestamps. I just made #711.

I assume the github oauth token is from Ralph's account, so I hardcoded `GH_USER=rhc54`. Regardless this has to be updated or not, we might want to encrypt this alltogether with `GH_TOKEN`.
Because of travis security rules, Ralph might also have to be the author of this PR before merging it.

rhc54:
I think, but could be wrong, that you won't need the ```GH_USER``` once this is committed into the actual repo as Travis is executed as me (since it is my token). Dunno about authorship - we can just try it as-is and see if it barks.

I have no idea how to test it other than commit and see what happens. Note that we already lost some of the pages as reported in #705. No idea why.

rhc54:
Kewl - I'd say let's give it a try. If all works okay, we can then add the ```---force``` option as I can't see any reason for keeping the ```gh-pages``` branch history. We can always regenerate the docs for a particular hash if someone wants to do so.

PR #711, opened 2018-08-31
Only deploy gh-pages when something did change

ggouaillardet:
Since the generated latex files have some timestamps,
manually update the gh-pages branch only if something has
changed in the docs/html directory.

The script/deploy.sh script does some extra steps (e.g. git clone)
in order to preserve the history on the gh-pages branches.

If we do not care about history, this step can be skipped and we can simply

git init; git push --force ...

note the GH_USER variable is currently hardcoded, and it might have to/could
be changed and/or encrypted (with GH_TOKEN)

Signed-off-by: Gilles Gouaillardet <gilles@rist.or.jp>


jsquyres:
FWIW: `--quiet` implies `--exit-code`.

PR #712, opened 2018-08-31
added --build-arg installplugins to control installing Jenkins plugins

beardeddog:
Default is not to install the plugins, this was chosen to reduce the chances that the docker build will fail because down stream repo where the plugins live is inaccessible (we've seen this).  You can then install these plugins later via the Jenkins console.

If you want to install the plugins, just add a `--build-arg installplugins=yes`


beardeddog:
For some reason this pull request included the three files from my previous pull request ... not sure how that happened or how to fix it or if it is a problem.

rhc54:
Try doing a "git rebase master" on your branch - it just needs updating to reflect that those have already gone in.

rhc54:
You may have to update your master branch first as well

PR #718, opened 2018-09-10
Added environment variable to override cmdline arguments

beardeddog:
Added the environment variable MTT_ARGS to set/override command line arguments.
This can be used in a couple ways,
  1) to reduce the number of command line arguments, or
  2) to override command line arguments already set in script invocations of pymtt

This request originated from inside Intel

Signed-off-by: beardeddog <beardeddogstudios@gmail.com>

### DIFF
--- a/pyclient/pymtt.py
+++ b/pyclient/pymtt.py
@@ -28,7 +28,14 @@ import argparse
 #    sys.exit('MTT usage error: add -h for help')
 
 # define the cmd line arguments
-parser = argparse.ArgumentParser(description="usage: %prog [options] testfile1 testfile2 ...")
+parser = argparse.ArgumentParser(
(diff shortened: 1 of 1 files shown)


pyclient/pymtt.py
```diff
@@ -28,7 +28,14 @@
 #    sys.exit('MTT usage error: add -h for help')
 
 # define the cmd line arguments
-parser = argparse.ArgumentParser(description="usage: %prog [options] testfile1 testfile2 ...")
```
ribab: Why remove the old description of MTT usage, instead of adding on to the description?

beardeddog: because its redundant with the usage that argparse provides

PR #721, opened 2018-09-18
Removed images files identified by protex scan

beardeddog:
Theses files were caught in the protex scan, three, background.png,
body-background.pg and hr.gif we don't use, bullet.png we can
live without and use just the default bullet indicator.


Jagaskak:
Jekyll Themes uses Creative Commons Zero v1.0 Universal Liscence. They don't have a header liscence to put in style.scss, but should it be mentioned somewhere?

beardeddog:
@Jagaskak ok ... let me see if I can make Protex believe it

beardeddog:
I was able to disposition these four files in protex and associate them with their create commons license, so will close this PR.  It might come back if I did it wrong, but for now it looks ok.

PR #723, opened 2018-09-19
Fix of ASIS Issue in MiddlewareBuild

### LINKED ISSUE #677: OMPI_Snapshot Plugin only works with ASIS enabled
The ompi_snapshot plugin fails when ASIS is not enabled with the following error:
```
2018-07-24 17:46:53.452164 LOGGING results for MiddlewareGet:OMPIMaster
2018-07-24 17:46:53.452185 =======================================
2018-07-24 17:46:53.452197 Exception was raised: <type 'exceptions.UnboundLocalError'> local variable 'status' referenced before assignment
2018-07-24 17:46:53.452207 =======================================
2018-07-24 17:46:53.452461 Traceback (most recent call last):

  File "/home/pgottesm/PyMTT/mtt/pylib/Tools/Executor/sequential.py", line 283, in execute
    plugin.execute(stageLog, keyvals, testDef)

  File "/home/pgottesm/PyMTT/mtt/pylib/Tools/Fetch/OMPI_Snapshot.py", line 211, in execute
    log['status'] = status

UnboundLocalError: local variable 'status' referenced before assignment
```
From a brief look over the code it looks like the log values are mostly set directly and the variable `status` is only set in one case, when ASIS is enabled. In all other cases, MTT will crash.

DebRez:
If a previous version_file is found matching a requested tarball exit the run. Previously OMPI_Snapshot was returning instead of exiting.

Fixes #677

### DIFF
--- a/pylib/Tools/Fetch/OMPI_Snapshot.py
+++ b/pylib/Tools/Fetch/OMPI_Snapshot.py
@@ -114,10 +114,11 @@ class OMPI_Snapshot(FetchMTTTool):
                         last_version = f.readline().strip()
                         f.close()
                         if last_version == snapshot_req.text.strip():
-                            log['status'] = 1
-                            log['stderr'] = "No new tarballs to test"
-                            # track that we serviced this one
-                            return
+                            status = 1
+                            stderr = "No new tarballs to test"
+                            stdout = None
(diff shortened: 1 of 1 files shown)


DebRez:
@PeterGottesman I will work on the problem some more and have you test the possible solutions before making another pull request.

PR #732, opened 2018-09-27
Remove perl client from mtt.spec

ribab:


ribab:
Because of ralph's change to remove perl client from this repo, we also need it removed from the spec file

PR #740, opened 2018-10-05
OMPI_Snapshot Plugin ASIS Fixed

DebRez:
When using version_file to prevent system from rerunning tests, MTT now
cleanly exits if the version_file is the same as the target tarball version.
self.options['version_file'] references the defaults only, changed to cmds['version_file'] to reflect parsed options.

Fixes # 677

### DIFF
--- a/pylib/Tools/Fetch/OMPI_Snapshot.py
+++ b/pylib/Tools/Fetch/OMPI_Snapshot.py
@@ -106,8 +106,9 @@ class OMPI_Snapshot(FetchMTTTool):
             sys.exit(-1)
 
         # check to see if we have already processed this tarball
+        # Fail and exit if requested version matches version file
         try:
-            if self.options['version_file'] is not None:
+            if cmds['version_file'] is not None:
                 if os.path.exists(cmds['version_file']):
                     try:
                         f = open(cmds['version_file'], 'r')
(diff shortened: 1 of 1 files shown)


pylib/Tools/Fetch/OMPI_Snapshot.py
```diff
@@ -106,23 +106,24 @@ def execute(self, log, keyvals, testDef):
...
                         last_version = f.readline().strip()
                         f.close()
                         if last_version == snapshot_req.text.strip():
                             log['status'] = 1
-                            log['stderr'] = "No new tarballs to test"
-                            # track that we serviced this one
-                            return
+                            log['stderr'] = "No new tarballs to test, exiting MTT"
+                            testDef.logger.verbose_print("Version file matches requested target, exiting MTT")
+                            sys.exit(1)
```
rhc54: This will cause the Python client to simply terminate, without any attempt at cleanup or posting the log (e.g., running the text reporter) so someone can see what happened. Is that really what you want to do? Wouldn't it be better to return an error code that indicates you shouldn't process any further dependencies and then let the logger run at the end?

FWIW: that's how the client currently executes, and it mirrors the Perl client in that behavior.

DebRez: Yes, I am going to close this pull request and rethink the issue.  Peter has his version_file symlinked to /dev/null.  I think that is actually the problem he is having, because I cannot reproduce his error any other way.  There are a couple of other bugs in OMPI_Snapshot to fix though so I will keep looking at it. What do you think? @rhc54

rhc54: Makes sense to me - I should think symlinking version_file to /dev/null would be forbidden? Not sure why that would make sense to do.

PeterGottesman: Sorry, I have not gotten a chance to look at this the past few days, should have a good chunk of time this weekend though. The issue is with attempting to use a version file which doesn't exist, or more specifically one where he data written in can't be read back out. This is of course why I used /dev/null, data written is won't be read back out. I went in to more detail in #677 about the actual cause I believe.

PR #764, opened 2018-11-15
Doxypypy for doxygen interepreter

ribab:
to get: pip install doxypypy

A more pythonic version of doxygen interpreter

Also, my system couldn't find old "doxypy.py" file


PeterGottesman:
JSYK, doxypy.py should have been provided by the required package doxypy in pyenv.txt. If it was completely removed the doxypy dependency should be replaced with doxypypy in that file.

PR #765, opened 2018-11-20
Add shell-mode to shell plugin

ribab:
Now you can run shell plugin just like you are running shell!

```
[TestRun:Example]
plugin = Shell
shell_mode = true
command = #!/bin/bash
          echo "hi"
          SOMEVAR=2
          echo somevar: $$SOMEVAR
          echo somevar+2: $$($$SOMEVAR + 2)

```

### DIFF
--- a/pylib/Tools/Build/Shell.py
+++ b/pylib/Tools/Build/Shell.py
@@ -51,6 +51,7 @@ class Shell(BuildMTTTool):
         self.options['allocate_cmd'] = (None, "Command to use for allocating nodes from the resource manager")
         self.options['deallocate_cmd'] = (None, "Command to use for deallocating nodes from the resource manager")
         self.options['asis_target'] = (None, "Specifies name of asis_target being built. This is used with \"ASIS\" keyword to determine whether to do anything.")
+        self.options['shell_mode'] = (False, "Use shlex or shell-mode for parsing?")
 
         self.allocated = False
         self.testDef = None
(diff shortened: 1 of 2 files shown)


DebRez:
Looks good to me.

pylib/Tools/Build/Shell.py
```diff
@@ -51,6 +51,7 @@ def __init__(self):
         self.options['allocate_cmd'] = (None, "Command to use for allocating nodes from the resource manager")
         self.options['deallocate_cmd'] = (None, "Command to use for deallocating nodes from the resource manager")
         self.options['asis_target'] = (None, "Specifies name of asis_target being built. This is used with \"ASIS\" keyword to determine whether to do anything.")
+        self.options['shell_mode'] = ("false", "Use shlex or shell-mode for parsing?")
```
PeterGottesman: Why not default to the boolean False instead of "false"? That way parseOptions will correctly translate the value to a boolean for you.

ribab: Hmm I didn't know about this parseOptions feature. I was using "lower()" to parse the options myself.

Are you saying that with boolean False, if I type "false" in the INI file it would still parse to be false as a boolean? I was thinking that it would give me a string "False" which would evaluate to True...

PeterGottesman: Yep, when you call parseOptions it calls `__convert_value` which converts based on the type of the default value.
https://github.com/open-mpi/mtt/blob/2451cb18676b7e31ea19829173f09fb1f5fb53d7/pylib/System/TestDef.py#L81-L91

ribab:
I added a BAT test, and I followed Peter's suggestion to use the __convert_option feature.

Ready for another review

PR #766, opened 2018-11-29
Fixes #744, removes --no-reporter, adds wildcard

### LINKED ISSUE #744: --no-reporter only works from the command line.
when I add 
export MTT_ARGS="--verbose --no-reporter"
to my .bash_profile only verbose works, data is sent to the reporter. LANL_DSR

beardeddog:
- removed --no-reported
- adds support for simple wildcards with --skip-sections
  wildcards in the form of
    - *text, text* and *text* will work
    - text*text will not becaus I just doing a simple
      substring search
    - this might work also with --sections, but I have
      not tested that
- fixed issue with spliting MTT_ARGS strings into commandline
  arguments, before it was doing a simple <space> split, now
  it looks for ' --'

### DIFF
--- a/pyclient/pymtt.py
+++ b/pyclient/pymtt.py
@@ -20,6 +20,7 @@ import logging
 import imp
 from yapsy.PluginManager import PluginManager
 import argparse
+import shlex
 
 # First check for bozo error - we need to be given
 # at least a cmd line option, so no params at all
@@ -103,9 +104,6 @@ execGroup.add_argument("-s", "--sections", dest="section",
                      help="Execute the specified SECTION (or comma-delimited list of SECTIONs)", metavar="SECTION")
 execGroup.add_argument("--skip-sections", dest="skipsections",
                      help="Skip the specified SECTION (or comma-delimited list of SECTIONs)", metavar="SECTION")
-execGroup.add_argument("--no-reporter", dest="reporter",
-                      action="store_true", default=False,
-                      help="Do not invoke any MTT Reporter modules")
 execGroup.add_argument("-l", "--log", dest="logfile", default=None,
                      help="Log all output to FILE (defaults to stdout)", metavar="FILE")
 execGroup.add_argument("--group-results", dest="submit_group_results", default=True,
@@ -158,7 +156,7 @@ args = parser.parse_args()
 # any set on the command line.  Environment will override the commandline.
 mttArgs = []
 if 'MTT_ARGS' in os.environ and os.environ.get('MTT_ARGS') != None:
-    mttArgs = os.environ['MTT_ARGS'].split()
+    mttArgs = shlex.split(os.environ['MTT_ARGS'])
     args = parser.parse_args(sys.argv[1:] + mttArgs)
 
 # check to see if MTT_HOME has been set - we require it
--- a/pylib/System/TestDef.py
+++ b/pylib/System/TestDef.py
@@ -517,6 +517,34 @@ class TestDef(object):
                     self.fill_log_interpolation("%s.%d" % (basestr, i), v)
         else:
             self.fill_log_interpolation(basestr, str(sublog))
+ 
+    def expandWildCards(self, sections):
+        expsec = []
+        cpsections = list(sections)
+        for sec in cpsections:
+            if '*' in sec:
+                modsec = sec.split('*')
+                startswith = modsec[0]
+                endswith = modsec[-1]
+                findsec = modsec[1:-1]
+                allsections = self.config.sections()
+                for s in allsections:
+                    if not s.startswith(startswith):
+                        continue
+                    if not s.endswith(endswith):
+                        continue
+                    found = True
+                    s_tmp = s
+                    for f in findsec:
+                        if not f in s_tmp:
+                            found = False
+                            break
+                        s_tmp = f.join(s_tmp.split(f)[1:])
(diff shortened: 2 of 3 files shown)


pyclient/pymtt.py
```diff
@@ -158,7 +155,8 @@
 # any set on the command line.  Environment will override the commandline.
 mttArgs = []
 if 'MTT_ARGS' in os.environ and os.environ.get('MTT_ARGS') != None:
-    mttArgs = os.environ['MTT_ARGS'].split()
+    # find all the ' --' command line args and split them
+    mttArgs = os.environ['MTT_ARGS'].replace(' --', ';--').split(';')
```
rhc54: Why would we do this? I'm not sure I understand the rationale.

pylib/System/TestDef.py
```diff
@@ -582,6 +582,20 @@ def configTest(self):
...
+        expsec = []
+        cpsections = list(sections)
+        for sec in cpsections:
+            if '*' in sec:
+                modsec = sec.replace('*','')
+                for s in self.config.sections():
+                    if modsec in s:
+                        expsec.append(s)
+                sections.remove(sec)
+        sections = sections + expsec
```
rhc54: not sure if it matters, but this does reorder what is in ```sections```

beardeddog: it does, it adds the expanded wildcard'd sections to the end of the sections list.

but for --skipsections it will not matter.  

Not sure if that will matter for --sections, but i think it won't matter either because we add based on the order of sections as generated from the ini file

pyclient/pymtt.py
beardeddog: here is an example, with MTT_ARGS set like
  `MTT_ARGS='--description=this is a test --skip-sections=Reporter:IUDatabase'`
we cannot split on space, but instead we need to split on something like ' --' to get each argument, but splitting 'eats' the characters you split on, so I change the ' --' to ';--' then split on ';' ...

ribab: maybe if you split using shlex?

`>>> shlex.split("--description='this is a test' --skip-sections=Reporter:IUDatabase")
['--description=this is a test', '--skip-sections=Reporter:IUDatabase']
`

pylib/System/TestDef.py
```diff
@@ -582,6 +582,21 @@ def configTest(self):
...
+        # this is simple wild carding, ie *text, text*, *text* and * 
+        # should all work, text*text will not
+        if sections is not None:
+            expsec = []
+            cpsections = list(sections)
+            for sec in cpsections:
+                if '*' in sec:
+                    modsec = sec.replace('*','')
+                    for s in self.config.sections():
+                        if modsec in s:
```
ribab: a simple replace and grep won't work in all cases.

For example:
> `--skip-sections='*skip'` won't work for sections:
> * TestRun:skip_some_other_section
> * TestRun:please_skip

Maybe instead, do this:

```
if sections is not None:
    expsec = []
    cpsections = list(sections)
    for sec in cpsections:
        if '*' in sec:
            modsec = sec.split('*')
            startswith = modsec[0]
            endswith = modsec[-1]
            findsec = modsec[1:-1]
            for s in allsections:
                if not s.startswith(startswith):
                    continue
                if not s.endswith(endswith):
                    continue
                found = True
                s_tmp = s
                for f in findsec:
                    if not f in s_tmp:
                        found = False
                        break
                    s_tmp = f.join(s_tmp.split(f)[1:])
                if not found:
                    continue
                expsec.append(s)
            sections.remove(sec)
    sections = sections + expsec
```

ribab: I tested this:

> rtbarell:~> python play.py
> filterSecs("*",['frontback', 'fronthi', 'hiback', 'fronthi2back', 'fronthi2', 'hi2back', 'hi2', 'hi'])
> ['frontback', 'fronthi', 'hiback', 'fronthi2back', 'fronthi2', 'hi2back', 'hi2', 'hi']
> 
> filterSecs("front*",['frontback', 'fronthi', 'hiback', 'fronthi2back', 'fronthi2', 'hi2back', 'hi2', 'hi'])
> ['frontback', 'fronthi', 'fronthi2back', 'fronthi2']
> 
> filterSecs("*back",['frontback', 'fronthi', 'hiback', 'fronthi2back', 'fronthi2', 'hi2back', 'hi2', 'hi'])
> ['frontback', 'hiback', 'fronthi2back', 'hi2back']
> 
> filterSecs("*hi2*",['frontback', 'fronthi', 'hiback', 'fronthi2back', 'fronthi2', 'hi2back', 'hi2', 'hi'])
> ['fronthi2back', 'fronthi2', 'hi2back', 'hi2']
> 
> filterSecs("*hi*back",['frontback', 'fronthi', 'hiback', 'fronthi2back', 'fronthi2', 'hi2back', 'hi2', 'hi'])
> ['hiback', 'fronthi2back', 'hi2back']
>

pyclient/pymtt.py
beardeddog: ok, I like this, it supports quoting a string with spaces, like the description which is more natural

pylib/System/TestDef.py
beardeddog: ok, I like this too

ribab:
Nice job testing your change

PR #779, opened 2019-01-04
Fix for using ompi with --no-shell and srun

DebRez:
Added support for ompi with --no-shell using srun in the SLURM.py
plugin. Tests were running on front end instead of compute nodes when
using salloc with --no-shell command with opmi.

### DIFF
--- a/pylib/Tools/Launcher/SLURM.py
+++ b/pylib/Tools/Launcher/SLURM.py
@@ -330,7 +330,7 @@ class SLURM(LauncherMTTTool):
         # get the "skip" exit status
         skipStatus = int(cmds['skipped'])
         # assemble the command
-        cmdargs = [cmds['command']].strip()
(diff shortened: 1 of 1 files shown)


ribab:
Hi @DebRez 

Sorry for not seeing this until now

I recommend against using --no-shell and allocate_cmd when possible...

One thing is that srun creates its own allocation.

I'm not sure if your change would have worked, but I do know that allocate_cmd with no-shell is not ideal because the allocation is not tied to your shell environment.

hppritcha:
@ribab the problem for us on large HPC systems such as at NERSC, if we were submitting individual unit tests with separate sruns that need to get an allocation, it would take days probably to run ~1000 tests.  For systems like cori at NERSC, its much more efficient to get a single SLURM allocation and then run the individual tests within this allocation.

PR #784, opened 2019-02-09
Added Copy plugin

noahv:
- Copies src to dest with the same behavior you would expect from cp
  - Copies directories recursively as well as individual files

### DIFF
--- a/pylib/Utilities/Copy.py
+++ b/pylib/Utilities/Copy.py
@@ -1,0 +1,100 @@
+from __future__ import print_function
+from builtins import str
+#!/usr/bin/env python
+#
+# Copyright (c) 2015-2019 Intel, Inc. All rights reserved.
+# $COPYRIGHT$
+#
+# Additional copyrights may follow
+#
+# $HEADER$
+#
+
+import shutil
+import distutils.dir_util
+import distutils.file_util
+import os
+from BaseMTTUtility import *
+
+from sys import version_info
+if version_info[0] == 3:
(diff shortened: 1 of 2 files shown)


pylib/Utilities/Copy.py
```diff
@@ -0,0 +1,99 @@
...
+from sys import version_info
+if version_info[0] == 3:
+  if version_info[1] <= 3:
+    from imp import reload
+  else :
+    from importlib import reload
+## @addtogroup Utilities
+# @{
+# @section Copy
+# Copy a Copy a comma separated list of files, directories, and links
```
rhc54: Very minor nit - "Copy a Copy a" typo

pylib/Utilities/Copy.py
```diff
@@ -0,0 +1,99 @@
...
+            pass
+        # perform the copy
+        try:
+            # Cleanup the target directory if it exists
+            if os.path.exists(dst):
+                shutil.rmtree(dst)
+            os.mkdir(dst)
+            for src in cmds['src'].split(','):
+                src = src.strip()
+                reload(distutils.dir_util)
```
rhc54: Haven't seen "reload" in other modules - is there some reason why it is required here? Perhaps a comment would be wise - just something to remind "future noah" as to the reason.

noahv: Added comment:

Clear the cache so that distutils.dir_util doesn't assume the same directory structure from last time things were copied

PR #789, opened 2019-02-14
configTest no longer  called between stages

ribab:
--clean is a part of configTest and shouldn't clean between stages

Now only ENV and LOG interpolation is refreshed between stages

### DIFF
--- a/pylib/System/TestDef.py
+++ b/pylib/System/TestDef.py
@@ -518,6 +518,50 @@ class TestDef(object):
         else:
             self.fill_log_interpolation(basestr, str(sublog))
 
+    def fill_env_hidden_section(self):
+        """fill ENV section with environment variables
+        """
+        try:
+            self.config.add_section('ENV')
+        except configparser.DuplicateSectionError:
+            pass
+        for k,v in os.environ.items():
+            self.config.set('ENV', k, v.replace("$","$$"))
+
+    def fill_log_hidden_section(self):
+        """Add LOG section filled with log results of stages
+        """
+        try:
+            self.config.add_section('LOG')
+        except configparser.DuplicateSectionError:
+            pass
+        thefulllog = self.logger.getLog(None)
+        for e in thefulllog:
+            self.fill_log_interpolation(e['section'].replace(":","_"), e)
+
+    def check_for_nondefined_env_variables(self):
+        # Check for ENV input
+        required_env = []
+        all_file_contents = []
+        for testFile in self.log['inifiles']:
+            file_contents = open(testFile, "r").read()
+            file_contents = "\n".join(["%s %d: %s" % (testFile.split("/")[-1],i,l) for i,l in enumerate(file_contents.split("\n")) if not l.lstrip().startswith("#")])
+            all_file_contents.append(file_contents)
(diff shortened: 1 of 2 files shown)


ribab:
I agree with you @rhc54 

The code needs comments and documentation in order to understand what it is doing.

I think it needs more than that though. So many features have been added into the backend of MTT overtime (many of those by me) which takes away from the modularity design of MTT. **Simple Backend. Extenable by plugins**.

I think an overhaul of MTT is needed to slim the backend... only issue is the time and resources to achieve that.

> I'm not sure I entirely understand but I have no major issues with these changes. One thing I note, however, is that the lack of comments makes it very hard to track what is going on. I therefore hope that there is some documentation somewhere that explains, for example, why you are scanning inifiles for what looks like some kind of $ENV symbol??? Is it written down somewhere why one would put such a thing in a file, what it means when you do so, etc? Because I have to confess: just trying to read this code leaves me with no earthly idea what it does.

To answer your question about `${ENV` being checked for in the code: A while ago, I enabled the ConfigParser to use its interpolation syntax, which allows variables to be read within an INI file. The variables are read as `${<SECTION>:<VARIBALE>}`. For example: `${MTTDefaults:description}` will return the description of the test provided in the INI file.

Taking advantage of this, I added two hidden sections. One called `ENV` and another called `LOG`.

`ENV` is filled with all the values in `os.environ`, which adds in a section similar to how an INI file would add one, but instead would look something like this:

    [ENV]
    HOME = /home/username
    PATH = /usr/bin:/usr/local/bin:/usr/bin/X11
    LANG=en_US.UTF-8
    ...

These values can then be accessed in an INI file like so:

    [TestRun:some_section]
    plugin = Shell
    command = echo my home is ${ENV:HOME}

The part of the code checking the keyword `${ENV` is checking if any environment variables a test uses are not defined before starting a test.

rhc54:
I guess that raises two questions:

1. How do we clean things up so we get back to a modular architecture?

1. Why did we diverge from modularity to begin with? Is there something lacking in the architecture? Or were we too focused on "just get it done" to think about the impact of what we were doing? Or...?

Recovering always takes longer than anyone would like as it has to be done in parallel with getting other work accomplished. Still, the problem will only grow worse if not addressed. Given you feel that many of these new features were added by you (and therefore you have a better idea of where the problems lie), would you mind putting together a brief summary of the problem? Just give us some idea of what was added that you feel negatively impacted modularity? It would help us to focus our thinking.

ribab:
> I guess that raises two questions:
> 
> 1. How do we clean things up so we get back to a modular architecture?
> 2. Why did we diverge from modularity to begin with? Is there something lacking in the architecture? Or were we too focused on "just get it done" to think about the impact of what we were doing? Or...?
> 
> Recovering always takes longer than anyone would like as it has to be done in parallel with getting other work accomplished. Still, the problem will only grow worse if not addressed. Given you feel that many of these new features were added by you (and therefore you have a better idea of where the problems lie), would you mind putting together a brief summary of the problem? Just give us some idea of what was added that you feel negatively impacted modularity? It would help us to focus our thinking.

I will write up a description of the problem :)

PR #796, opened 2019-02-26
Doxygen improvements

DebRez:
Used https://github.com/Feneric/doxypypy instructions to add a
filter pattern to the Doxygen file. Removed .ini files as part of
documentation because they were causing errors. Updated the environment
files to include doxypypy (Tried adding it to .travis.yml but got
package not found error). Made sure travis was using python 2.7 when
doing the doxygen build.

[Ticket: X]


DebRez:
Looks like this broke the Travis builds - working on it.

PR #812, opened 2019-03-23
Refactored Environment Module Support

noahv:
- New global TestDef structure maintains requested modules
  - Added checkForModules, applyModules, and revertModules helpers
  - Added tests/bat/Autotools sample autotools based hello world
  - Added tests/bat/ModuleCmd ini tests for plugins modified
  - Updated extensive_example.ini to include new modules_swap


DebRez:
@noahv In the Autotools.py the added return statements on line 304, 342, 379 and 403 will cause the program to fail because the position in the directory structure will not be correct - need to return to cwd before returning.

noahv:
Thanks Deb for the review!  I should have mentioned I was off on spring break backpacking!  I am back this week.

I will update the patch today.

Issue #816, opened 2019-05-28
Error getting IBM tests - python client


DebRez:
The last time I had a successful MTT run was with the ompi-master 05/20/19. Since then I am getting the error Exception was raised: <type 'exceptions.OSError'> [Errno 2] No such file or directory:    scratch/TestGet_IBM/ompi-tests/ibm

This is happening on different systems and I have made no changes to my .ini file.

I can git clone the ompi-tests repository manually and the IBM folder transfers over fine so it seems to be a problem with MTT.  Is anyone else seeing this problem?

DebRez:
Here is the TestGet section of my INI file
[ASIS TestGet:IBM]
plugin = Git
url =  git@github.com:open-mpi/ompi-tests
subdir = ibm
parent = MiddlewareBuild:OMPIMaster

DebRez:
Here is the output from the TestGet:
========================================================================
START EXECUTING [ASIS TestGet:IBM] start_time=2019-05-27 21:57:20.591756
========================================================================
OPTIONS FOR SECTION: ASIS TestGet:IBM
  plugin = Git
     url = git@github.com:open-mpi/ompi-tests
  subdir = ibm
  parent = MiddlewareBuild:OMPIMaster
Executing plugin Git
Git Execute
Working repo git@github.com:open-mpi/ompi-tests
Working final repo git@github.com:open-mpi/ompi-tests
ExecuteCmd start: git clone git@github.com:open-mpi/ompi-tests
stdout: Cloning into 'ompi-tests'...
stderr: Checking out files:   5% (600/10030) 
Checking out files: 100% (10030/10030), done.
ExecuteCmd done
Fetch CMD: merge_stdout_stderr stderr_save_lines pr scratchdir modules_unload pl
atform subdir branch pwfile asis username description parent time password url s
tdout_save_lines modules trial modules_swap executor organization
LOGGING results for ASIS TestGet:IBM
DONE EXECUTING [ASIS TestGet:IBM] end_time=2019-05-27 21:57:53.415520, elapsed=0:00:32.823764
STATUS OF [ASIS TestGet:IBM] IS 0

ASIS TestBuild:IBMInstalled

===================================================================================
START EXECUTING [ASIS TestBuild:IBMInstalled] start_time=2019-05-27 21:57:53.415800
===================================================================================
OPTIONS FOR SECTION: ASIS TestBuild:IBMInstalled
               parent = TestGet:IBM
  merge_stdout_stderr = 1
[...]

PR #818, opened 2019-05-29
Revert before autotools commit that broke Autotools.py

ribab:
https://github.com/open-mpi/mtt/issues/816

The previous commit  https://github.com/open-mpi/mtt/commit/f649e180595cf1f66de9156929dff16b3ffcc0d6 broke Autotools, so this is to revert that commit.

The previous commit did not fix what it intended to fix, so reverting the commit.


DebRez:
Great, thanks.

PR #823, opened 2019-06-28
Add new Fetch plugins for rpms and Python pkgs

rhc54:


ribab:
The plugins look great, but question about the PIP plugin -- for a test, wouldn't it be better to have a virtualenv plugin? 

Also, pip install has a "--user" flag which installs it locally to the user, which might be better in some cases than installing it for everyone.

rhc54:
> The plugins look great, but question about the PIP plugin -- for a test, wouldn't it be better to have a virtualenv plugin?

We could support both, I imagine - the user may or may not be running MTT in a virtualenv. What needs to be changed in this plugin to enable it to work both ways? I had just assumed that pip would install into the virtualenv if that is where it was being run - does it need a special flag to do so?

> Also, pip install has a "--user" flag which installs it locally to the user, which might be better in some cases than installing it for everyone.

Agreed - will add that flag.

rhc54:
Okay, I finally realized what you meant by the virtualenv comment. I'll add a separate plugin for that option.

PR #827, opened 2019-07-03
Update Git fetch tool to include subdir in location

rhc54:
If someone specified a subdirectory in their ini file stage to fetch a
Git repo, then the location should be updated with that subdir.


rhc54:
@ribab @beardeddog @noahv  @DebRez @hppritcha 

Note that this changes the location reported by the Git plugin when the ini file specifies a "subdir". We previously weren't returning the "subdir" in the location, which isn't what the ini file specifically requested. Please check to ensure your ini files are updated accordingly.

PR #828, opened 2019-07-03
 Update PMIx unit test runner to cover python tests

rhc54:
Specify the command to use to run tests so that C-based unit tests can
utilize their starter while Python unit tests can just execute. Update
PYTHONPATH is Python libraries were installed by middleware. Allow
specification of middleware to be used for tests.


ribab:
Looks good

PR #829, opened 2019-07-12
Update Docker README

dsikich:
This updates the Docker README on setting up a cherrypy server
with a postgres DB. Using the current instructions results
in errors when making server requests. These turn on port
forwarding when running cherrypy and postgres.

### DIFF
--- a/server/docker/README.docker.txt
+++ b/server/docker/README.docker.txt
@@ -4,14 +4,26 @@ docker network create -d bridge --subnet 172.31.0.0/24 docker_network
 # Build postgres docker image
 docker build --build-arg http_proxy=$http_proxy --build-arg https_proxy=$https_proxy -t mttdb-postgres -f Dockerfile.postgres .
 # Start the postgres server
-docker run --name mttdb-postgres --network=docker_network --ip=172.31.0.2 -e POSTGRES_PASSWORD=mtt -d mttdb-postgres 
+docker run --name mttdb-postgres --network=docker_network -p 5432:5432 -e POSTGRES_PASSWORD=mtt -d mttdb-postgres 
 # Test container
-PGPASSWORD="mtt" psql -a --host=172.31.0.2 --dbname=mtt --username=postgres
+PGPASSWORD="mtt" psql -a --host=localhost --dbname=mtt --username=postgres
 
+# port forwarding for -p N:N may be different. This uses port forwarding on a local machine to point to the docker container.
+
+defined to use port forwarding on a local machine
 # Build cherrypy docker image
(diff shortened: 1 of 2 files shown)


server/docker/README.docker.txt
```diff
@@ -4,14 +4,23 @@ docker network create -d bridge --subnet 172.31.0.0/24 docker_network
 # Build postgres docker image
 docker build --build-arg http_proxy=$http_proxy --build-arg https_proxy=$https_proxy -t mttdb-postgres -f Dockerfile.postgres .
 # Start the postgres server
-docker run --name mttdb-postgres --network=docker_network --ip=172.31.0.2 -e POSTGRES_PASSWORD=mtt -d mttdb-postgres 
+docker run --name mttdb-postgres --network=docker_network --ip=172.31.0.2 -p 5432:5432 -e POSTGRES_PASSWORD=mtt -d mttdb-postgres
```
rhc54: I am no expert by any means, but I noted that you changed the IP address to "localhost" or "0.0.0.0" elsewhere - do you need to change the 172.31.0.2 address here as well?

dsikich: @rhc54 yeah, well I just tested not specifying the ip and just using the -p option, and that works fine. I don't think the ip specified there is doing anything. I will take that out.

rhc54: Kewl - does the "5432" number correlate to a port forwarding you defined separately? Just want to ensure that we captured all the steps here.

dsikich: @rhc54 yeah that is the only place it is specified

rhc54: Okay, could you please add a comment in there somewhere indicating that the `-p N:N` should be replaced with whatever port forwarding was defined - perhaps include an example of the command? Otherwise, someone might think that is a "magic" number.

dsikich: not sure I fully understand, example of which command? The run line? Or do you mean point out that the port forwarding in that line may not be the same, like:

docker run --name mttdb-postgres -p N:N -e POSTGRES_PASSWORD=mtt -d mttdb-postgres

where -p N:N is the port forwarding that was defined?

rhc54: Yes - I gather you defined a port forwarding on your local machine to point the port at the docker container. I just want that noted somewhere in this file so someone reading it five years from now realizes that they need to forward a port and use that value in that command line. Otherwise, they are likely to wind up smacking their head after thrashing for an hour 😄

dsikich: Makes sense, I added a comment before running the commands

PR #830, opened 2019-07-13
Revise ExecuteCmd return value and Refactor Launchers

rhc54:
Every time we want to extend the return values from ExecuteCmd, we have
to go back and change every place we call it. This isn't supportable in
the long term. So instead of passing back arrays of values, return a
dictionary instead. This makes it possible to extend the return values
from ExecuteCmd without having to modify any existing code.

Sadly, it did mean having to modify everyone who currently uses that
plugin. Thus, this PR touches a number of plugins - but fortunately
doesn't alter their logic.

The primary driver for the changes was the need to refactor the various
Launchers. There was a ton of copy/paste'd code in them - turned out
that the differences were actually rather small. So instead of having
all that code duplication, put the common code in the LauncherMTT stage
plugin from which all the Launcher's derive.

Reason for doing this is that there were some significant problems in
the existing launchers:

* the timeout support was never implemented, and so we didn't detect or
properly handle tests that simply hung/spun forever

* the designation of skipped vs failed tests was incorrect. I think we
got a little too clever in coding things to the minimum number of lines
of code.

* using the return status as an indicator of "skipped" and friends meant
that we were losing the actual exit code of the test.

* we incorrectly included the #skipped tests in the #tests being
executed, thus causing us to execute fewer tests than the user had
specified

This has now been fixed. The logic for creating the lists of skipped and
expected-to-fail tests is all centralized and broken into distinct
steps. Yes, it is a few more lines of code - but hopefully it is a
little easier to read and understand.

Executing the tests is likewise centralized. Each launcher is
responsible for just creating its cmd line for executing the tests.
Everything else is just done with calls to the common code.

Still up: updating the reporters to properly understand the new tag
designators for test results (skipped, failed as expected, etc.)

### DIFF
--- a/pylib/Tools/Harasser/Harasser.py
+++ b/pylib/Tools/Harasser/Harasser.py
@@ -85,7 +85,7 @@ class Harasser(HarasserMTTTool):
     def parallel_execute(self, cmds, cmdargs, testDef):
         """This function is passed into multiprocessing as a target
         """
-        status,stdout,stderr,time,_ = testDef.execmd.execute(cmds, cmdargs, testDef)
+        results = testDef.execmd.execute(cmds, cmdargs, testDef)
 
     def start(self, testDef):
         """Harassment is started on the system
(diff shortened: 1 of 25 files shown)


rhc54:
Hi folks - I have updated both the TextFile and IUDatabase reporter to properly handle the tagging of test results as pass, fail, skip, or timed out. I couldn't test the IUDatabase plugin, so I would appreciate someone doing so. The Textfile reporter is working correctly.

pylib/Tools/Harasser/Harasser.py
```diff
@@ -85,7 +85,7 @@ def get_running_harassers(self):
     def parallel_execute(self, cmds, cmdargs, testDef):
         """This function is passed into multiprocessing as a target
         """
-        status,stdout,stderr,time = testDef.execmd.execute(cmds, cmdargs, testDef)
+        status,stdout,stderr,time,_ = testDef.execmd.execute(cmds, cmdargs, testDef)
```
hppritcha: I'm not a python expert, but is this correct?  the return value isn't being treated as a dictionary like in other places in the PR.

rhc54: Garrrr - missed that one. Fixed it. Thanks!

PR #838, opened 2019-07-22
Use a different method for timeout detection

### LINKED ISSUE #837: subprocess.TimeoutExpired only works with Python3
https://github.com/open-mpi/mtt/blob/a4f291390f442e537a15445688f11a3852f1afa7/pylib/Utilities/ExecuteCmd.py#L149

subprocess.TimeoutExpired in ExecuteCmd.py doesn't work with Python2.

I know that Python2 isn't supported anymore by the Python community, but are we still planning on supporting Python2?

rhc54:
Apparently, Python 2 does not have the subprocess "timeout" feature in
it. One solution would be to have everyone install the subprocess32
module that backported such features to Python 2, but that would be
intrusive. So instead use a threaded model for tracking timeout.

Little more cumbersome and involves another thread, but appears to be
working just fine under test.

Fixes #837 

### DIFF
--- a/pyclient/debug.ini
+++ b/pyclient/debug.ini
@@ -1,0 +1,108 @@
+#Copyright (c) 2018-2019 Intel, Inc.  All rights reserved.
+#
+
+# Set defaults
+[MTTDefaults]
+scratch = mttscratch
+description = Run Server/Client test
+platform = Intel Bend
+executor = sequential
+
+# Get the system profile
+[Profile:Installed]
+
+#======================================================================
+# Build PMIx dependencies: libevent, libev, hwloc, and cython
+#======================================================================
+[ASIS MiddlewareGet:libevent]
+plugin = FetchTarball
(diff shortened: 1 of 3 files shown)


rhc54:
@ribab Seems to be something going on with Travis as we are hanging on a "gcc --version" command.

However, I hadn't remembered that we do test Python 2.7 in Travis. Since we do test that environment, and the current code (pre this PR) passed all Travis tests - I don't understand issue #837. Is the Travis Python 2.7 environment different from the one you are using?

ribab:
I don't understand why TravisCI passes with the python2 version

When I create a new python2.7 environment on my machine and I import subprocess, I don't see subprocess.TimeoutExpired defined

rhc54:
It turns out that it isn't defined in Python 2.7. My guess is that the Travis installation includes subprocess32, which backported timeout support (among other things) from Python 3.

It appears I found the magic fix while working on the Intel machines, so even the Travis 2.7 is now happy with this method...and I have verified it works on the Intel side. Hooray!

PR #851, opened 2019-08-12
Fix import error

DebRez:
MTT failing due to missing import shlex statement in
mtt/pylib/Tools/Launcher/LauncherMTTTool.py.

### DIFF
--- a/pylib/Stages/Profile/DefaultProfile.py
+++ b/pylib/Stages/Profile/DefaultProfile.py
@@ -65,10 +65,15 @@ class DefaultProfile(ProfileMTTStage):
         cmds = {}
         testDef.parseOptions(log, self.options, keyvals, cmds)
         keys = list(cmds.keys())
-        opts = self.options.keys()
+        # pass in a timeout option as not every system will support
+        # every option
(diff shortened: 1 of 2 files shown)


rhc54:
Looks like a stalled Python 2.7 build - restarted it

rhc54:
There is clearly a bug in the default profile plugin. I honestly don't know how it ever worked this way, but since it wasn't passed any options it should have never executed anything. Yet the output shows it is executing at least `uname -v` and then appears to enter an infinite loop. I'll try to fix it and add it to your branch since you are seeing it

rhc54:
My bad - the plugin should default to running all defined options with default values. However, it appears that some of those commands occasionally can "hang" on the Travis system under Python 2.7, thus causing it to fail. So add some protection - let's see if that can pass

rhc54:
There is something strange going on between Travis and the Mac - we are getting quite different behavior when parsing options and then cycling across the resulting list. This caused the code I was locally testing to pass every time, but fail every time on Travis. Not sure what to make of it, but something we will need to keep an eye upon.

Issue #852, opened 2019-08-12
Most recent version of MTT failing on three different systems.


DebRez:
MTT is failing on Fog, Trinitite and Cori systems since the most recent version was pulled. One clear error was the missing shlex import in LauncherMTTTool.py.

Pull request #851 is meant to fix this issue but the travis build is failing. The only change made was to add the missing "import shlex" statement. Can you help troubleshoot this @rhc54 ?

rhc54:
Sure - can you give me any more info on how it is failing?

DebRez:
I have seen several issues. The missing shlex import, ERROR (stderr) No tests found, and ERROR (stderr) Too many return codes were supplied to satisfy the number of tests expected to fail: 2:1

Here is the relevant parts of my INI file:
                 fail_tests = abort final
                 fail_timeout = max_procs
                 fail_returncodes = 1, 3

rhc54:
How long has it been since you updated? Just want to get some idea of the change involved as those areas of the code were modified like a month or more ago, AFAIK

DebRez:
I just updated last thursday.

rhc54:
I was asking how long it had been before that 😄

DebRez:
Many months

rhc54:
wow - that might not be a good idea as it can let problems go by until it is very hard to track them down. If you are going to update that infrequently, then you probably should create some CI or nightly regression so we have a chance to catch them earlier on.

I take it those are Cray machines? What Python version are you using?

DebRez:
python 2.7. Trinitite  and Cori are cray systems, Fog is a "commodity" system - all intel hardware.

rhc54:
Is there anything common about those machines - e.g., they are all running under Slurm?

Can you post the full .ini section that is failing? I'd like to see what plugin is being used and the full set of options. Best I can tell: it would appear your option parser may be getting confused. Note that right now the LaunchMTTTool ignores the "fail_timeout" option - doesn't hurt to have it, but won't do anything

DebRez:
I am getting the error "No tests found" now and it looks like the only place that line shows up is in  LauncherMTTTool.py. 

Those machines are all running under Slurm. Here is the .ini file for Fog:

[LauncherDefaults:OMPI]
plugin = SLURM
command = srun
options = -n 16 -c 4 -t 15:00
skipped = 77
merge_stdout_stderr = 1
stdout_save_lines = 100
stderr_save_lines = 100

[...]

PR #853, opened 2019-08-13
Be more flexible in how we take input

rhc54:
Accept lists of tests expected to fail and tests to skip that are
delimited by commas, spaces, or tabs. Instead of having a separate list
of expected return codes for failed tests, specify those in the same
place where we list those tests:

fail_tests = ftest1:3 ftest2 ftest3:-5

where we specify the expected return code with a ":code" suffix to the
name of the test. Tests without an expected return code will be marked
as "passed" if they fail regardless of what they return.

Add a "dryrun" option to the IUDatabase reporter so we can see what it
would have sent without actually sending anything to the db.

Refs #852 

### DIFF
--- a/pylib/Stages/Reporter/IUDatabase.py
+++ b/pylib/Stages/Reporter/IUDatabase.py
@@ -40,6 +40,7 @@ requests.packages.urllib3.disable_warnings(InsecureRequestWarning)
 # @param keep_debug_files   Retain reporter debug output after execution
 # @param debug_server       Ask the server to return its debug output as well
 # @param email              Email to which errors are to be sent
+# @param dryrun             Print debug without actually submitting to database server
 # @}
 class IUDatabase(ReporterMTTStage):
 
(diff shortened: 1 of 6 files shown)


rhc54:
@DebRez I modified the input processing code for failed tests to be more flexible - it now accepts comma, space, and tab delimiters (and any mix of them). I tested it all with the SLURM launcher plugin and the IUDatabase reporter and everything was working for me. Please update your .ini file and give it a try

rhc54:
@ribab @beardeddog 

NOTE: this makes a change to the way we process .ini files for failed tests. If you have any ini files that use that option, they will require a change when this is committed.

rhc54:
> I suggest editing all references to fail_tests to reflect the new usage model. Something like:
> self.options['fail_tests'] = (None, "Names of tests that are expected to fail. Can use space or comma between entries. Include the expected return code using the following format: test_name:#")

Done!

rhc54:
@DebRez @ribab Any last comments/changes?

PR #861, opened 2019-09-03
Sets np correctly

### LINKED ISSUE #852: Most recent version of MTT failing on three different systems.

DebRez:
Determines np before tests are run then assigns np correctly in LauncherMTTTools.py.

Fixes #852

### DIFF
--- a/pylib/Tools/Launcher/LauncherMTTTool.py
+++ b/pylib/Tools/Launcher/LauncherMTTTool.py
@@ -465,12 +465,15 @@ class LauncherMTTTool(IPlugin):
                     testLog['result'] = testDef.MTT_TEST_PASSED
                     self.numPass += 1
             try:
-                testLog['np'] = cmds['np']
+                testLog['np'] = log['np']
             except KeyError:
                 try:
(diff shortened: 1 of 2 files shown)


rhc54:
I'm afraid this doesn't make a lot of sense to me. If I'm following it correctly, you are dealing with some Slurm cmd options that implicitly specify the number of procs to run, but aren't including an explicit `-np` option. I get that, but putting the `-np` in the log and extracting it later seems weird and risky as you have now put a value in the final overall log, not the log of that particular test.

Why not just update `cmds["-np"]` with the number of procs you calculate? The rest of the logic would then flow pretty naturally and the value would only go into the correct log.

DebRez:
I will do that. Will have to touch a lot more of @ribab 's code which I was avoiding doing.

PR #866, opened 2019-10-10
swat undefined variable name bug

hppritcha:


noahv:
Just looking into why we didn't hit this previously...

Issue #867, opened 2019-10-22
IPMITool plugin type error


beardeddog:
running the IPMITool plugin with python3, got this error

```
Executing plugin IPMITool
IPMITool execute
IPMITool: target controller username password pwfile command maxtries numthreads dryrun sudo modules_unload modules modules_swap trial scratchdir description platform organization merge_stdout_stderr stdout_save_lines stderr_save_lines executor time
=======================================
Exception was raised: <class 'TypeError'> '<' not supported between instances of 'int' and 'tuple'
=======================================
Traceback (most recent call last):

  File "/opt/mtt/pylib/Tools/Executor/sequential.py", line 290, in execute_sections
    plugin.execute(stageLog, keyvals, testDef)

  File "/opt/mtt/pylib/Tools/CNC/IPMITool.py", line 303, in execute
    if len(targets) < self.options['numthreads']:

TypeError: '<' not supported between instances of 'int' and 'tuple'
```

beardeddog:
this  has been fixed

PR #868, opened 2019-10-22
#867 fixed TypeError in IPMITool plugin

beardeddog:
#867 fixed TypeError in IPMITool plugin

### DIFF
--- a/pylib/Tools/CNC/IPMITool.py
+++ b/pylib/Tools/CNC/IPMITool.py
@@ -300,10 +300,10 @@ class IPMITool(CNCMTTTool):
         self.threads = []
         # release the queue
         self.lock.release()
-        if len(targets) < self.options['numthreads']:
+        if len(targets) < cmds['numthreads']:
             rng = len(targets)
         else:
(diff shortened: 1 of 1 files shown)


pylib/Tools/CNC/IPMITool.py
```diff
@@ -300,7 +300,7 @@ def execute(self, log, keyvals, testDef):
         self.threads = []
         # release the queue
         self.lock.release()
-        if len(targets) < self.options['numthreads']:
```
ribab: correct way is using `cmds['numthreads']`. -- `cmds` should have the final value for numthreads

pylib/Tools/CNC/IPMITool.py
```diff
@@ -300,7 +300,7 @@ def execute(self, log, keyvals, testDef):
...
+        if len(targets) < self.options['numthreads'][0]:
             rng = len(targets)
         else:
             rng = self.options['numthreads']
```
ribab: Also, this line should be changed to `cmds['numthreads']`

PR #870, opened 2019-10-25
ELK-friendly log output

ribab:
just add --elk flag, and output will change to a more elk-friendly
format

* functionality to redirect all prints to testDef.logger.print
* event logging for
    * each stage
    * each execmd command
    * all options
    * all envvars

### DIFF
--- a/pylib/Utilities/Logger.py
+++ b/pylib/Utilities/Logger.py
@@ -13,6 +13,8 @@ from builtins import str
 import sys
 import datetime
 from BaseMTTUtility import *
+import json
+import os
 
 ## @addtogroup Utilities
 # @{
@@ -30,6 +32,9 @@ class Logger(BaseMTTUtility):
         self.cmdtimestamp = False
         self.timestampeverything = False
         self.stage_start = {}
+        self.elk = None
+        self.current_section = None
+        self.execmds_stash = []
 
     def reset(self):
         self.results = []
@@ -41,7 +46,7 @@ class Logger(BaseMTTUtility):
     def print_options(self, testDef, prefix):
         lines = testDef.printOptions(self.options)
         for line in lines:
-            print(prefix + line)
+            self.print(prefix + line)
         return
 
     def open(self, testDef):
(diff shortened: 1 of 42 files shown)


pylib/Utilities/Logger.py
```diff
@@ -133,8 +159,76 @@ def stage_end_print(self, stagename, log):
...
+                elklog['other']['compiler_version'] = log['compiler']['version']
+            for k in log:
+                if k != 'options' and k != 'parameters' \
+                and k != 'time_start' and k != 'time_end' and k != 'time' \
+                and k not in elklog and k not in elklog['other']:
+                    elklog['other'][k] = str(log[k])
+            print('SECTION', json.dumps(elklog))
+        self.current_section = None
+
+#{'compiler': {'compiler': 'gnu', 'status': 0, 'version': 'gcc (GCC) 8.3.0'},
```
noahv: Did you intend to add these commented out fields

ribab: nope, those were for reference. I'll delete them

beardeddog:
1. will the EXEC_CMD have start and end times?
2. can we id the order they were executed?
3. they probably should be inside the section they were run .. too

`EXEC_CMD {"section": "Profile:Installed", "cmdargs": ["uname", "-r"], "status": 0, "stdout": ["4.12.14-195-default"], "stderr": [], "timedout": null, "starttime": null, "endtime": null, "elapsed": null}`

ribab:
> 1. will the EXEC_CMD have start and end times?
> 2. can we id the order they were executed?
> 3. they probably should be inside the section they were run .. too
> 
> `EXEC_CMD {"section": "Profile:Installed", "cmdargs": ["uname", "-r"], "status": 0, "stdout": ["4.12.14-195-default"], "stderr": [], "timedout": null, "starttime": null, "endtime": null, "elapsed": null}`

1. EXEC_CMD start and end times are currently set by the plugin, and depends on some other variable on whether they are set. I will experiment with unconditionally sending these to the elk-output
2. That's a good idea. I will see if I can track the execution order with an id
3. The section is included in the EXEC_CMD json, **if that isn't enough** I can figure out how to save these json outputs and include them as a list with the SECTION json output

ribab:
This pull request might be abandoned in favor of https://github.com/open-mpi/mtt/pull/874

PR #871, opened 2019-10-29
Capture slurm jobid in ExecuteCmd

ribab:
Currently, it is difficult to match a run of MTT with a job in the slurm logs.

What this change makes possible is:

1. run MTT and get slurm job ids in output
2. use `scontrol show job <SLURM JOBID>` to see the slurm log info for that job

This is specific to slurm, so any other kid of job ids to capture will need this functionality to be extended.

Signed-off-by: Barella, Richard T <richard.t.barella@intel.com>


ribab:
I added some comments

PR #880, opened 2019-12-05
Limit modifying slurm jobname to only if we're running "srun"

ribab:
srun doesn't show any slurm jobid in its stdout, so we need to modify
SLURM jobid only for srun

sbatch and salloc both output the slurm jobid into the stdout so we can
still capture it there.

This fixes a bug where MTT is overriding the slurm jobname when a user of MTT expects it to not change when using "allocate_cmd" and "deallocate_cmd"

This also fixes a bug in shell.py:
> `if 'stderr' in log:`
> is correct instead of
> `if log['stderr']:`

### DIFF
--- a/pylib/Utilities/ExecuteCmd.py
+++ b/pylib/Utilities/ExecuteCmd.py
@@ -169,7 +169,7 @@ class ExecuteCmd(BaseMTTUtility):
         except:
             pass
 
-        return slurm_jobids
+        return list(set(slurm_jobids))
 
 
     def execute(self, options, cmdargs, testDef, quiet=False):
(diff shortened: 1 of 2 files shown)


pylib/Utilities/ExecuteCmd.py
```diff
@@ -209,7 +212,7 @@ def execute(self, options, cmdargs, testDef, quiet=False):
                 continue
             arg = arg.replace('\"','')
             # Look for any job names in cmdargs, and attach unique identifier
-            if arg.startswith('--job-name='):
+            if cmdargs[0] == 'srun' and arg.startswith('--job-name='):
```
noahv: "-J" is an alternate form of --job-name

ribab: I take care of that in a later line in which I already am checking for `srun`

https://github.com/ribab/mtt/blob/644f9a92271bb5b239ad4e9a494ea1ed773f011d/pylib/Utilities/ExecuteCmd.py#L219

PR #882, opened 2019-12-06
Fix annoying mpi_get_version output

ribab:
By using `|sort` before `|uniq -c`, and by removing the extra newline after the mpi version output, **mpi_get_verion** output is reduced to only a few lines.

### DIFF
--- a/pylib/Utilities/MPIVersion.py
+++ b/pylib/Utilities/MPIVersion.py
@@ -101,7 +101,7 @@ int main(int argc, char **argv) {
     char version[3000];
     int resultlen;
     MPI_Get_library_version(version, &resultlen);
-    printf("%s\\n", version);
+    printf("%s", version);
     return 0;
 }""")
             fh.close()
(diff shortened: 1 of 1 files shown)


rhc54:
You might want to add a "-n 1" to the mpiexec line and the equivalent to the others. The command as-written will execute ```mpi_get_version``` as many times as there are allocated slots - probably where some of that annoying duplication is coming from.

ribab:
True, but I think it's possible some compute nodes accidentally have a different MPI version installed.

rhc54:
FWIW: the various MPI implementers spent some effort to keep at least the "-n" option portable. It would be a really weird implementation that didn't support it - I very much doubt you have one installed anywhere.

ribab:
Oh that's not what I meant by different versions.

I mean that mpi_get_version prints the current MPI version that is running, and when that script is ran on every compute node, some compute nodes might have a different version.

`mpiexec ./mpi_get_version |sort |uniq -c`
If there are 32 compute nodes this command will output the version of MPI for 32 compute nodes. If one compute node version isn't the same as the others that will be apparent in the output.

PR #883, opened 2019-12-06
updated ELK output to remove arrays

beardeddog:

### DIFF
--- a/pylib/Utilities/Logger.py
+++ b/pylib/Utilities/Logger.py
@@ -35,6 +35,8 @@ class Logger(BaseMTTUtility):
         self.timestampeverything = False
         self.stage_start = {}
         self.elk_id = None
+        self.elk_caseid = None
+        self.elk_cycleid = None
         self.elk_head = None
         self.elk_log = None
         self.current_section = None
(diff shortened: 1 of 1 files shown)


pylib/Utilities/Logger.py
```diff
@@ -161,7 +170,7 @@ def log_to_elk(self, result, logtype):
                         uid = pwd.getpwnam(user).pw_uid
                         gid = grp.getgrnam(group).gr_gid
                         os.chown(allpath, uid, gid)
-            self.elk_log = open(os.path.join(self.elk_head, '{}.elog'.format(self.elk_id)), 'a+')
+            self.elk_log = open(os.path.join(self.elk_head, '{}-{}.elog'.format(self.elk_caseid, self.elk_id)), 'a+')
```
ribab: Nice catch

pylib/Utilities/Logger.py
```diff
@@ -119,34 +125,37 @@ def log_to_elk(self, result, logtype):
...
                             or isinstance(v, datetime.date) else v)
                   for k,v in result.items()}
         result['execid'] = self.elk_id
+        result['cycleid'] = self.elk_cycleid
+        result['caseid'] = self.elk_caseid
         result['logtype'] = logtype
+
+        # drop the stderr and stdout, will use the copies that are part of the commands issues
+        if 'stderr' in result:
+            del result['stderr']
```
ribab: If MTT captures an exception, maybe due to an error in Python, that exception is stored in `result['stderr']`

The `stderr` and `stdout` in the commands array is only populated from ExecuteCmd plugin and doesn't include when a syntax error happens within another plugin in MTT.

beardeddog: the observed behavior is that the last command issued is also populating this outer `stderr` and `stdout` which is why I dropped it

ribab: I think with your pull request, we should merge it since it works, and then we can do some testing to see if all the errors we care about make it into ELK

PR #886, opened 2019-12-12
moving code from python2 to python3

beardeddog:
For this change I used the 3.6.8 version of 2to3 to detect and make the
bulk of the changes.  I also changed all #!/usr/bin/env python to
python3


beardeddog:
I should also say that I ran all of our BAT tests against the changes in the PR and the all passed with no issues

PR #887, opened 2019-12-13
Add elk-friendly output options in argparse options

ribab:
to better document usage instead of hiding behind os.environ


beardeddog:
can we move the ELK options into their own group, ie they are in debugGroup right now, how about an ELKGroup ?

ribab:
@beardeddog sounds good

> can we move the ELK options into their own group, ie they are in debugGroup right now, how about an ELKGroup ?

Fixed this

PR #892, opened 2020-01-07
Improved LOG-interpolation syntax and made plugin to debug using that syntax

ribab:
This syntax is something like ${LOG:TestRun_mystage.status} which enables you to access
the internal MTT log from an INI file. This is a powerful feature which I have been able
to use to even make an INI file that tests itself.

Includes enhances to documentation regarding `${ENV: ...}` and `{LOG: ...}` interpolation syntax

### DIFF
--- a/docs/pages/user_guide.md
+++ b/docs/pages/user_guide.md
@@ -382,3 +382,61 @@ password = database password
 platform = Your_Platform  
 url = https://mtt.open-mpi.org/submit/cpy/api/  
 
+## Capabilities of MTT
+
+### Accessing environment variables from within an INI file
+
+One useful feature of MTT is the ability of accessing outside environment variables from within an INI file. This enables MTT tests to respond to dynamic input fron the environment.
+
+To access environment variables, use this syntax:
+
+```
+${ENV:ENVIRONMENT_VARIABLE}
+```
+
+MTT will directly substitute the value into the location of the INI file you specified.
+
+If an environment variable does not exist, MTT will warn the user and MTT will exit with an error.
+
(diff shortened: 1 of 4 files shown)


rhc54:
Only one concern: we seem headed down the path that drove the perl client as we add more and more obtuse, non-obvious semantics to create new behavior. Just like with the perl client, we seem destined to have so much complexity that nobody truly understands what all this stuff does, nor can anyone really maintain it. Just too much really specific functionality that is driven by hard-to-grok syntax.

I'm not sure of the answer here, but I am getting increasingly concerned that the code (and all its functionality) may already be non-maintainable, or will soon get there. It makes me wonder if maybe we are all so driven to make immediate progress on minor things that we aren't properly looking at the best way to use (or modify) the architecture,or even asking folks about the best way to achieve a given goal, but instead just "patch" something onto something else so we can make it work and move on to other things.

Not picking on this one in particular - it is something I've been noticing for quite some time and getting increasingly concerned about. As I said, I don't have an immediate solution to propose - everyone is busy with their own deadlines. I just fear that the Python client will suffer the same fate as its Perl cousin in another year or two - used by a couple of "gurus" and otherwise avoided at all costs...and heaven help the fool who commits a change to it!

ribab:
Hi @rhc54 , I hear and agree with your concern, however I am merging this PR since it doesn't apply to this PR specifically.

rhc54:
Actually, I think it does apply to this PR - I just meant I wasn't singling it out. Still, go ahead and merge it. As we discussed on the call today, we need to go back and clean up things like this so we don't leave a legacy of such problems.

PR #893, opened 2020-01-14
Moved ElkLogger functionality into its own module

ribab:
This is to reduce confusion when maintaining Logger.py

There is also a minor fix to a race condition in creating the elog files which uses `pathlib.Path` to create directories and not error out out if they already exist.

### DIFF
--- a/pylib/Utilities/ElkLogger.py
+++ b/pylib/Utilities/ElkLogger.py
@@ -15,6 +15,7 @@ import datetime
 from BaseMTTUtility import *
 import json
 import os
+from pathlib import Path
 import pwd
 import grp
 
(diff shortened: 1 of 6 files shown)


pylib/Utilities/ElkLogger.py
```diff
@@ -0,0 +1,132 @@
...
+import os
+from pathlib import Path
+import pwd
+import grp
+
+## @addtogroup Utilities
+# @{
+# @section ElkLogger
+# Log results to *.elog files with JSON entries to be later consumed by Elastic Stack (ELK)
+# Uses pymtt.py options: elk_head, elk_id, elk_testcycle, elk_testcase, elk_debug, elk_chown, elk_maxsize, elk_nostdout, elk_nostderr
```
beardeddog: can we explain what each of these are used for?

beardeddog: elk_head, elk_id, elk_testcycle, elk_testcase, elk_debug, elk_chown, elk_maxsize, elk_nostdout, elk_nostderr  env vars that is

PR #910, opened 2020-12-04
Adjust max date check to 2050. Long live MTT.

jjhursey:


ribab (approved):
Not that I understand why there is a year check like this in the first place, but the change looks simple enough

jsquyres (approved):

PR #911, opened 2020-12-05
Script to prune the DB to a sliding window

jjhursey:
* Takes one argument the number of months to keep in the DB

### DIFF
--- a/server/sql/support/prune-db-window.pl
+++ b/server/sql/support/prune-db-window.pl
@@ -1,0 +1,296 @@
+#!/usr/bin/env perl
+
+#
+# Josh Hursey
+#
+# Usage:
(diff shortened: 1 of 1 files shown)


jjhursey:
I have this script running in a cron. I'd like to keep an eye on it for a couple of months before merging this PR just in case there is anything we need to tweek. For now it looks pretty good.

So I'll leave it in draft form until it's proven itself a bit more.

jsquyres (approved):
image

jjhursey:
All the other maintenance scripts are in Perl so I'm just keeping the theme.

jsquyres:
jen-yes

jjhursey:
This script has been working great. Merging so we don't lose track of it.

PR #913, opened 2021-02-19
Added check for slurm job step abort time out

beardeddog:
This covers a situation where the job is aborted but the slurm cleanup times out, leaving
the job status as 0, here is the typical output

     stderr: srun: Force Terminated job 6050
     stderr: srun: Job step aborted: Waiting up to 32 seconds for job step to finish.
     stderr: srun: error: Timed out waiting for job step to complete

A dict was added to the slurm SLURM class to contain an error string to search for and
results and test return codes.  This is passed to the LauncherMTTTool, runTests function
where this additional check can be made.  The code is structured this way to allow the
SLURM plugin to do these checks and allow other launcher plugins to not do these checks.

### DIFF
--- a/pylib/Tools/Launcher/SLURM.py
+++ b/pylib/Tools/Launcher/SLURM.py
@@ -45,8 +45,18 @@ import subprocess
 class SLURM(LauncherMTTTool):
 
     def __init__(self):
+        # This covers a situation where the job is aborted but the slurm cleanup times out, leaving
+        # the job status as 0, here is the typical output
+        #     stderr: srun: Force Terminated job 6050
+        #     stderr: srun: Job step aborted: Waiting up to 32 seconds for job step to finish.
+        #     stderr: srun: error: Timed out waiting for job step to complete
+        self.falsePass = {}
+        self.falsePass['errstr']  = "Timed out waiting for job step to complete"
+        self.falsePass['rtncode'] = 143
+        self.falsePass['result']  = 0     # testDef.MTT_TEST_FAILED
+
         # initialise parent class
-        LauncherMTTTool.__init__(self)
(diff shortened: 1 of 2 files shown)


pylib/Tools/Launcher/SLURM.py
```diff
@@ -45,7 +45,18 @@
...
+        #     stderr: srun: Force Terminated job 6050
+        #     stderr: srun: Job step aborted: Waiting up to 32 seconds for job step to finish.
+        #     stderr: srun: error: Timed out waiting for job step to complete
+        self.falsePass = {}
+        self.falsePass['errstr']  = "Timed out waiting for job step to complete"
+        self.falsePass['rtncode'] = 143
+        self.falsePass['result']  = 0     # testDef.MTT_TEST_FAILED
+
         # initialise parent class
+        # LauncherMTTTool.__init__(self, additionalCheck = self.falsePass)
```
noahv, line 59: Did you leave this in intentionally?

noahv (commented):
Looks good thanks Bill

pylib/Tools/Launcher/SLURM.py
beardeddog: thanks, that's a testing artifact, the commented line is the one that should be uncommented ... none of these changes will take effect with out that line enabled.  Good catch.

beardeddog: i've uploaded the fix

noahv (commented):
I am wondering if this should be conditional .ini file switch to set the string to search for because of the added time required to parse output logs for SLURM plugin users that don't expect any force terminated jobs. 

On the other hand it is a good thing to always try and catch false positives...

ribab (approved):
The changes look good to me

Issue #914, opened 2021-03-12
cherrypy server code experiences traumas moving from python2 to python3


hppritcha:
The cherrypy related server code has a number of python2isms that caused the installation at mtt.open-mpi.org to start failing once that system was upgraded and python2 was no longer supported.

hppritcha:
Actually turns out @DebRez had fixed many of the python3 problems in cherrypy but we'd not updated mtt on the mtt.openmpi.org install recently.  did that today.  still need PR #915

hppritcha:
closed via #915 and refresh of mtt on mtt.open-mpi.org

ribab:
How did the update go @hppritcha ?

PR #915, opened 2021-03-12
cherrypy: update base64 call for python3

hppritcha:
related to issue #914

Signed-off-by: Howard Pritchard <hppritcha@gmail.com>


jjhursey (approved):
Good catch!

PR #918, opened 2021-03-23
IUDatabase reporter - fix a bunch of problems

hppritcha:
The IUDatabase scheme for reporting testing results to the cherrypy server
had some major flaws.  It would not report anything back to the cherrypy server if

- there was a failure to build the middleware
- there was a failure to build tests

Only if the run got through starting at least a few tests would it actually report anything.

This PR reworks the way that the results of a MTT run are reported so that results are
reported to the cherrypy server regardless of whether the middleware build fails, or the
test build(s) fail.

Signed-off-by: Howard Pritchard <hppritcha@gmail.com>


jjhursey (approved):
I'm a bit rusty with the data structures in the python client, but this looks ok to me.

PR #933, opened 2022-04-08
Fix usage of 'exec' key in AlreadyInstalled plugin

devreal:
The check trying to find the `exec` used the wrong dictionary (`keyval` instead of `cmds`) and didn't check for whether `exec` was set in the first place.

At a larger scale, the documentation [1] does not mention `exec` (or the `module_unload` key) and it's not clear to me what that check is actually for. Isn't it sufficient to check whether `command` specified in `LauncherDefaults` is available?

[1] https://open-mpi.github.io/mtt/pages/user_guide.html#ini-section-middlewareget

Signed-off-by: Joseph Schuchart <schuchart@icl.utk.edu>

### DIFF
--- a/pylib/Tools/Fetch/AlreadyInstalled.py
+++ b/pylib/Tools/Fetch/AlreadyInstalled.py
@@ -61,6 +61,10 @@ class AlreadyInstalled(FetchMTTTool):
         # parse any provided options - these will override the defaults
         cmds = {}
         testDef.parseOptions(log, self.options, keyvals, cmds)
+
+        if cmds['exec'] is None:
+            log['status'] = 0
+            return
   
         # Apply any requested environment module settings
         status,stdout,stderr = testDef.modcmd.applyModules(log['section'], cmds, testDef)
@@ -71,7 +75,7 @@ class AlreadyInstalled(FetchMTTTool):
             return
 
         # now look for the executable in our path
-        if not find_executable(keyvals['exec']):
+        if not find_executable(cmds['exec']):
             log['status'] = 1
             log['stderr'] = "Executable " + cmds['exec'] + " not found"
         else:


hppritcha (approved):

hppritcha:
@devreal I could see why one might want to have command in launcher be something like say srun, whereas if one's building tests etc, it might be a useful sanity check to see if the compiler/script (aka mpicc) is in the path.

The docs need a general audit.

ribab:
This is a good check to see if exec is present - but shouldn't it be an error if it's not present? Maybe setting `stderr` with the error string and returning `1` instead of `0`?

hppritcha:
good point, i'd vote for returning an error.

devreal:
Why? It's neither documented nor necessary for the rest of the MTT execution. If I don't need to check whether the exec is available (because I know it is) then why do you make me define it nevertheless? And then guess what caused the error if I don't? I consider this an entirely optional key so it should not cause an error if it's not provided. If you want to make it a required key and have a good reason then it should be documented first.

hppritcha:
I'm okay with making this field optional.  @ribab what do you think?

ribab:
The purpose of the AlreadyInstalled plugin is check if there is an executable in the path. This field is documented

`self.options['exec'] = (None, "Executable that should be in path")`

If this field isn't included then there is no purpose in using the plugin.

I guess it doesn't really matter whether the plugin returns 0 or 1 because it is a no-op plugin anyways

devreal:
> The purpose of the AlreadyInstalled plugin is check if there is an executable in the path. This field is documented
> 
> `self.options['exec'] = (None, "Executable that should be in path")`

That's not documentation if I have to skim through the code to figure out why my run fails. And funny thing: all the other (modules related) fields default to `None` and are not required. 

There is a use-case besides checking for the executable: the combinatorial executor requires a `MiddlewareGet` section to work properly (don't know why, it's not documented either). That's why I used it and came across this issue here.

Issue #891, opened 2019-12-30
Error when not including any middleware stages when using Combinatorial Executor


ribab:
Error when not including any middlware stages when using Combinatorial Executor

Below is the error

    ExecuteCombinatorial
    Error, empty run log, combinatorial executor failed

After only adding a MiddlwareGet stage, the error goes away.

ribab:
This line is where the error happens:
https://github.com/open-mpi/mtt/blob/fd44e12f39af3b7b01fea0f557d74c1c86bae6e3/pylib/Tools/Executor/combinatorial.py#L175

PR #934, opened 2022-04-08
Combinatorial: warn if TestRun or MiddlewareGet sections are missing

devreal:
The Combinatorial executor would fail with a cryptic error saying that no run log is available. Be a little nicer and at least say what's missing...

Addresses https://github.com/open-mpi/mtt/issues/891 (not sure it's a real fix, just playing a little nicer)

Signed-off-by: Joseph Schuchart <schuchart@icl.utk.edu>


ribab:
I'm not opposed to this change. Have you tested that this fails with the better error message when either TestRun or MiddlewareGet is missing?

devreal:
I did, it throws the warning that I added if either of this sections are missing.

ribab:
Merging

PR #935, opened 2022-04-08
sequential_exec: conditionally chdir

hppritcha:
if the parent has set a location in is logfile.
This may not always be the case.

### DIFF
--- a/pylib/Tools/Executor/sequential.py
+++ b/pylib/Tools/Executor/sequential.py
@@ -292,14 +292,17 @@ class SequentialEx(ExecutorMTTTool):
                         parent = None
                     if parent:
                         parent_log = testDef.logger.getLog(parent)
-                        parent_loc = parent_log['location']
+                        parent_loc = None
+                        if 'location' in parent_log:
+                            parent_loc = parent_log['location']
                     if checking_run_if:
                         testDef.logger.verbose_print("run_if command: %s" % run_if_cmd)
                         if run_if_cmd:
                             cmdargs = shlex.split(run_if_cmd)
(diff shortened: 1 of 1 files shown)


pylib/Tools/Executor/sequential.py
```diff
@@ -292,14 +292,16 @@ def execute_sections(self, testDef):
...
+                        if 'location' in parent_log:
+                            parent_loc = parent_log['location']
                     if checking_run_if:
                         testDef.logger.verbose_print("run_if command: %s" % run_if_cmd)
                         if run_if_cmd:
                             cmdargs = shlex.split(run_if_cmd)
                             if parent:
                                 original_loc = os.getcwd()
-                                os.chdir(parent_loc)
+                                if parent_loc is not None:
```
ribab, line 303: parent_loc would not be defined. this will throw an error.

an example: https://www.online-python.com/PTYOAEbRrg

ribab (changes requested):

pylib/Tools/Executor/sequential.py
hppritcha: nice catch...

hppritcha: @ribab check now

hppritcha:
@ribab check now

pylib/Tools/Executor/sequential.py
ribab: I think that fixes it

ribab (approved):

ribab:
approved

PR #941, opened 2022-12-15
Add --enable-mpirun-prefix-by-default to sample config

sunkuamzn:
Prevents test failures caused by ORTE not setting the prefix correctly

Signed-off-by: Sai Sunku <sunkusa@amazon.com>


sunkuamzn:
I see `--enable-mpirun-prefix-by-default` option when I type `./configure -h` on the nightly 4.1.x tarball. But I can switch to `--enable-orterun-prefix-by-default` if that's more backwards compatible

hppritcha:
I didn't realize the mpirun-prefix-by-default was available in older versions.  in that case ignore my comment.  i'll merge this PR

PR #943, opened 2023-01-09
Changes from Open MPI AWS server

jjhursey:
* Use `env` to get the `python` library
 * Disable the `access.log` file. It doesn't provide any useful information.
 * Make `get_cursor` function more robust
 * Handle some UTF-8 characters
 * Adjust a few log messages from `debug` to `info`

Signed-off-by: Joshua Hursey <jhursey@us.ibm.com>


jjhursey:
This is the current diff from the Open MPI AWS instance. I figured we should push it upstream just in case the server encounters some data loss, and it might be helpful to others.

hppritcha (approved):

PR #945, opened 2023-03-09
Clone a separate repo for each TestGet section

wenduwan:
This change alters the git clone behavior for TestGet sections.

Currently, if TestGet:XXX and TestGet:YYY section requires the same git repository, the repo is only cloned once, which is unlikely the desired behavior.

With this change, each TestGet section will have a separate clone of the repo. This prevents dependent TestRun conflicts.

### DIFF
--- a/pylib/Tools/Fetch/Git.py
+++ b/pylib/Tools/Fetch/Git.py
@@ -148,6 +148,9 @@ class Git(FetchMTTTool):
         if repo.endswith(".git"):
             repo = repo[0:len(repo)-4]
 
+        dst = os.path.join(testDef.options['scratchdir'], log['section'].replace(":","_"))
+        repo = os.path.join(dst, repo)
+
         # Apply any requested environment module settings
         status,stdout,stderr = testDef.modcmd.applyModules(log['section'], cmds, testDef)
         if 0 != status:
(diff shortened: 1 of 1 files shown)


hppritcha:
would it be possible to post an ini file which works correctly with this change but fails with the current head of master?

wenduwan:
> would it be possible to post an ini file which works correctly with this change but fails with the current head of master?

Yes sure. Here is an example based on [this example](https://github.com/open-mpi/mtt/blob/master/samples/python/mpitest.ini)

```
...
[TestGet:IBM]
parent = MiddlewareBuild:OMPIMaster
plugin = Git
url =  https://github.com/open-mpi/ompi-tests
username = rhc54
pwfile = rhcpasswd
subdir = ibm

[TestGet:Intel]
parent = MiddlewareBuild:OMPIMaster
plugin = Git
url =  https://github.com/wenduwan/ompi-tests # <- User wants to use a fork
subdir = intel_tests
...
```

HEAD currently will first process `TestGet:IBM` and cache `https://github.com/open-mpi/ompi-tests` result in `self.done['ompi-tests']`, and skip `https://github.com/wenduwan/ompi-tests` in `TestGet:Intel`, because `os.path.basename(urlparse(url).path)` only returns the repo basename `ompi-tests`.

Based on my understanding of the design intention, the desired behavior is to separate the repo for each `TestGet:XXX`, i.e.
```
TestGet_IBM/ompi-tests/imb
TestGet_Intel/ompi-tests/intel_tests
```

